PR #1, opened 2023-03-01
Implement Kotlin with Jetpack bootcamp task

schuetzcarl:
Not convinced by the Jetpack Framework, I would suggest to not merging it after the review exercise. But let's see together


DanielTavaresA (commented):
Unit tests verify well the desired functionnalities. Code is well written. LGTM

schuetzcarl:
For the moment, we decided to not use JetPack Compose Framework

PR #2, opened 2023-03-01
Bootcamp/firebase

jules552:
Implemented the liaison between client and Firebase and created a small activity with one test to try the feature

I think we should merge since we're gonna obviously use a database and that's create a very basic template of how to use it (and also provide the different dependencies needed to use it)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -2,6 +2,7 @@ plugins {
     id 'com.android.application'
     id 'org.jetbrains.kotlin.android'
     id 'jacoco'
+    id 'com.google.gms.google-services'
 }
 
 android {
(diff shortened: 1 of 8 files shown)


schuetzcarl (approved):
Thanks for your work !

I agree with you, and think that it should be merged.

I just had the following small remarks on it :

Why casting emailView and phoneView when declared in get() and set() methods ?
Why is the @JvmStatic tag needed in the test class companion object ?

Some cosmetic stuff :
	In FirebaseActivityTest, Import androidx.test.espresso.Espresso.onView to not prefixing it every time in setAndGetTest()
	Remove the unusued imports in that same class
	In FirebaseActivity, "dbRef" or "dbRootRef" may be more informative than only "db" for the DatabaseReference class attribute

jules552:
I updated and taking account major part of your comments; 

I just prefer to keep the @JvmStatic since it works and it's how I've seen people using it when doing my research.
And I kept the "db" name since there is only one db in it's also how they name the variable in the google's examples website so

PR #3, opened 2023-03-01
Bootcamp/maps

yanvow:
### Add the usage of Google Maps

- add a MapsMarkerActivity and its xml file
- open the map activity after clicking on the validate button
- display a map centered at EPFL coordinates
- add controls and gestures to the map
- add zoom buttons to the map
- add a marker at satellite's location
- display a toast when the sat marker is clicked with a fun message and sat's coordinates


TimVW7 (commented):
Looks good. We could merge if we needed this code. Not many tests were run, but it seems to be by design with the part that was implemented (can't really be tested).

PR #4, opened 2023-03-02
Bootcamp/authentication

DanielTavaresA:
Code for authentication with google account using firebase. Some changes would be necessary but could be added to the project


jules552:
Thank you for you work,

The general code seems pretty clean, some unused imports need to be cleaned up but other than that I think the code is pretty safe explanatory and doesn't require that much comments.
One small issue is the hardcoded strings you use, you should not use hardcoded string since it will be complicated to have multiple languages supported for example.
Of course the only big issue is for the moment the lack of tests so it needs to be fixed before eventually merging the code.

PR #23, opened 2023-03-11
Feature/proof photo

jules552:
So here the activity for the moment, you can link the fragment to anything for example a button, it will open you camera app, you take a picture and the picture is send to firebase storage if connected, if not, just keep the local file generated (later maybe next sprint, adding a subroutine to check if the file is not empty, if that's the case, try sending the file to firebase storage)

No big issue known, just Google Camera can bug sometime during the test or usage, but not my fault honestly.

I modified the ci script also since there was no camera included, so I added the option for emulated one.

I don't think what i've made should brake anything in the main, I just think i'm not on the must up to date version (sorry), but only my 2 files Fragment/Test should be added and what got configured, the rest is not important.


jules552:
A bit better coverage, can't do really more since the rest is just the condition when we ask for permissions but testing permissions is kinda not useful and also not really possible since revoking permissions for each test is not possible and lead to crash cf. https://stackoverflow.com/questions/43462172/android-revoke-permission-at-start-of-each-test

schuetzcarl (changes requested):
Note that I had an Exception when testing manually on my local emulator.
It happens whenever I don't set any rights prior to the launch and click on : 
- "While using the app" 
or
- "Only this time" 
in the "Allow Factotum to take pictures and record video" screen.

In your tests classes, I remarked constant waiting time values, are they chosen from a manual testing. It may be the reason why the CI fails sometimes.

Otherwise thank you for your super clean and well commented code. 
I agree with merging it after those little fixes.

jules552:
So i've used so voodoo sorcery to fix issue with permission since activitylauncher should be created before the fragment from what i have understood, but no idea why it was not crashing when the permission was given when it should have...
I've added the time in the name of the photo and rename JPG with USER, when the authentification will be finish we could hopefuly replace this by the actual name of the user.
I remove unacessary import and also the tag, mainly used for debugging.

schuetzcarl (approved):
Great work, thank you.

I agree for merging it.

PR #24, opened 2023-03-12
Feature/maps activity

DanielTavaresA:
- Adds a UI to select dummy routes to be shown
- Adds a map UI that shows the destinations of the routes selected (= in the future would be client destinations)
- Adds a option to show all the destinations from the list
- Adds an option to generate the bike itinerary in google maps app

NB: Problem with Cirrus CI because Google Services work only with API 31+. Therefore CI is constantly failing on tests linked to the maps API because a screen shows the error : "won't run without google play services which are not supported by your device" (can be seen on recordings of the tests). Still not found a way to resolve this issue.


app/src/androidTest/java/com/github/factotum_sdp/factotum/RouteTest.kt
schuetzcarl, line 1: As RouteTest.kt is not an AndroidTest class, I would suggest to move it to its counterpart package, which as to be created :
app/src/**test**/java/com/github/factotum_sdp/factotum/ui/

schuetzcarl (changes requested):
Overall clean work, with a welcomed effort on the documentation. The ViewModel pattern is perfectly managed. Thank you for that work !

As soon as the minor remarks about the code have been resolved, I will agree to merge that PR.

app/src/androidTest/java/com/github/factotum_sdp/factotum/RouteTest.kt
DanielTavaresA: Not sure to understand why RouteTest.kt should be in a UI package. It is more a data oriented class than UI. Could you clarify your idea ?

schuetzcarl: Sorry you are right, its counter package is indeed the data package which has to be created :

app/src/**test**/java/com/github/factotum_sdp/factotum/data

schuetzcarl (approved):
Thank you for the merge :)

PR #25, opened 2023-03-14
Feature/contact diary

TimVW7:
Implemented main UI structure for the list of contacts. Meaning a list of contacts one can scroll through with the option to select any contact to see details of the contact (address, phone number, etc...). Design still needs work but at least the basics are implemented.


jules552:
Overall, really great job.

The main issue for me is that you did not really place what you did in the "directory" section as intented by Carl I think, so what you did is not linked at all to the other parts of the project.

Of course the layout and  mainly how the profil are displayed are not too esthetic but that's a issue for next sprints I think.

So other than that, the comments and the code seems understandable and clear to me. The tests seems to cover all the parts of what you did.

I'm in favor of merging this just after modifying this only major issue but that should not me too long. (If that was in fact the intention of  this directory menu in the main activity)

TimVW7:
Put the code into the directory section of the code. Should be good for merge now.

jules552 (approved):
It's perfect, and I agree for merging now !

PR #20, opened 2023-03-09
Feature/login UI

yanvow:
Implement the login UI:

- login fragment
- buttons to navigate to other fragments (e.g. Sign Up fragment, Main fragment)
- text fileds to enter email & password
- google One Tap authentication option

PR #27, opened 2023-03-15
Feature/login UI merge

yanvow:
Adapted from #20


app/src/main/java/com/github/factotum_sdp/factotum/data/LoginDataSource.kt
```diff
@@ -0,0 +1,24 @@
...
+            // TODO: handle loggedInUser authentication
+            val fakeUser = LoggedInUser(java.util.UUID.randomUUID().toString(), username)
+            return Result.Success(fakeUser)
+        } catch (e: Throwable) {
+            return Result.Error(IOException("Error logging in", e))
+        }
+    }
+
+    fun logout() {
+        // TODO: revoke authentication
```
TimVW7, line 22: Logout and revoking authentification seems like 2 different processes, which one are you trying to implement here ?

app/src/main/java/com/github/factotum_sdp/factotum/data/LoginRepository.kt
```diff
@@ -0,0 +1,46 @@
...
+        // @see https://developer.android.com/training/articles/keystore
+        user = null
+    }
+
+    fun logout() {
+        user = null
+        dataSource.logout()
+    }
+
+    fun login(username: String, password: String): Result<LoggedInUser> {
```
TimVW7, line 30: Nitpicking here, but it makes more sense in my opinion to have the login function before the logout function.

app/src/main/java/com/github/factotum_sdp/factotum/data/LoginDataSource.kt
yanvow: We are probably going to use Firebase's methods to logout.
These methods probaly revoke the authentication to log you out

app/src/main/java/com/github/factotum_sdp/factotum/ui/login/LoginFormState.kt
```diff
@@ -0,0 +1,10 @@
+package com.github.factotum_sdp.factotum.ui.login
+
+/**
+ * Data validation state of the login form.
```
TimVW7, line 4: If you could reformulate your explanation for this class it would be great, I have some trouble understanding what it is going to be used for simply from it's description/code.

TimVW7 (changes requested):
Overall very nice, modular code, well written. Please add more comments to make the code more digestible. And your code should test more of the basic features of each fo your classes. One thing that might be good for future sprints, try to divide up the task in multiple smaller pull requests so that it is easier to keep track of everything.

PR #34, opened 2023-03-16
Better settings for CI and adding Firebase Emulator

jules552:
Faster compilation for CI using 8 cores (around -2min) and adding firebase emulator correctly configured.

We should merge it.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,6 +1,6 @@
 container:
   image: cirrusci/android-sdk:30
-  cpu: 4
+  cpu: 8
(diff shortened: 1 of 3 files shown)


.cirrus.yml
```diff
@@ -1,6 +1,6 @@
 container:
   image: cirrusci/android-sdk:30
-  cpu: 4
+  cpu: 8
```
yanvow, line 3: Very nice idea.

yanvow (approved):
Nice initiative

PR #46, opened 2023-03-17
End/first sprint

schuetzcarl:
Adjust work to be merged before the 1st Sprint presentation meeting. 
Coming from #27

PR #47, opened 2023-03-17
Feature/user roles merge

yanvow:
Add the notion of role in login
Each user has a role that is either :
- Boss
- Courier
- Client


schuetzcarl:
Thank you @yanvow for fixing your PR and the Issues with the CI. 
However some of the code could belongs to the future #7 issue, and are not traversed yet, that's why I created a new PR with a selection proposal. I let you agree or not with that, and add your unit tests for the back-end.

See #46

PR #58, opened 2023-03-20
center the map on all displayed markers

yanvow:
Added a private function to add the markers on the map and center the camera on them
(forgot to name the branch feature/...)

### DIFF
--- a/app/src/main/java/com/github/factotum_sdp/factotum/ui/maps/MapsFragment.kt
+++ b/app/src/main/java/com/github/factotum_sdp/factotum/ui/maps/MapsFragment.kt
@@ -22,6 +22,7 @@ import com.google.android.gms.maps.model.LatLngBounds
 class MapsFragment : Fragment() {
 
     companion object {
+        private val EPFL_LOC = LatLng(46.520536, 6.568318)
         private const val ZOOM_PADDING = 100
         private const val minZoom = 6.0f
         private const val maxZoom = 14.0f
@@ -29,6 +30,7 @@ class MapsFragment : Fragment() {
 
     private var _binding: FragmentMapsBinding? = null
     private val viewModel: MapsViewModel by activityViewModels()
+
     // This property is only valid between onCreateView and
     // onDestroyView.
     private val binding get() = _binding!!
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/factotum_sdp/factotum/ui/maps/MapsFragment.kt
```diff
@@ -68,6 +67,20 @@ class MapsFragment : Fragment() {
...
 
+    private fun placeMarkers(routes : MutableList<Route>, googleMap: GoogleMap){
+        val bounds = LatLngBounds.Builder()
+
+        for (route in routes){
+            route.addDstToMap(googleMap)
+            bounds.include(route.dst)
+        }
+
+        val padding = 100 // offset from edges of the map in pixels
```
DanielTavaresA, line 78: Could be turned to a constant

app/src/main/java/com/github/factotum_sdp/factotum/ui/maps/MapsFragment.kt
```diff
@@ -50,11 +52,8 @@ class MapsFragment : Fragment() {
         mapFragment.getMapAsync{ googleMap ->
             // clears map from previous markers
             googleMap.clear()
-            //Shows destinations of selected roads
-            for (route in viewModel.routes.value.orEmpty()){
-                route.addDstToMap(googleMap)
-            }
-            googleMap.moveCamera(CameraUpdateFactory.newLatLng(EPFL_LOC))
+            // places markers on the map and centers the camera
+            placeMarkers(viewModel.routes.value!!, googleMap)
```
DanielTavaresA, line 56: Route values could be null and should be handled instead of throwing an exception. Maybe could just check if null before, if not place the Markers.

DanielTavaresA (changes requested):
Overall good code. Some tests would be welcome to check if functionnalities are well implemented

DanielTavaresA (changes requested):

DanielTavaresA (approved):
Corrections have been rightly made. LGTM

PR #59, opened 2023-03-20
Add tests to increase code coverage

yanvow:
Add tests for the login UI to increase code coverage

### DIFF
--- a/app/src/main/java/com/github/factotum_sdp/factotum/MainActivity.kt
+++ b/app/src/main/java/com/github/factotum_sdp/factotum/MainActivity.kt
@@ -7,7 +7,9 @@ import androidx.appcompat.app.AppCompatActivity
 import androidx.drawerlayout.widget.DrawerLayout
 import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.findViewTreeLifecycleOwner
+import androidx.navigation.Navigation
 import androidx.navigation.findNavController
+import androidx.navigation.fragment.NavHostFragment
 import androidx.navigation.ui.AppBarConfiguration
 import androidx.navigation.ui.navigateUp
 import androidx.navigation.ui.setupActionBarWithNavController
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/github/factotum_sdp/factotum/MainActivity.kt
```diff
@@ -34,7 +36,9 @@ class MainActivity : AppCompatActivity() {
 
         val drawerLayout: DrawerLayout = binding.drawerLayout
         val navView: NavigationView = binding.navView
-        val navController = findNavController(R.id.nav_host_fragment_content_main)
```
schuetzcarl, line 37: I don't get why it's a refactor here ? It seems that these two new lines does the job of one.

app/src/main/res/layout/content_main.xml
```diff
@@ -6,7 +6,7 @@
     app:layout_behavior="@string/appbar_scrolling_view_behavior"
     >
 
-    <fragment
+    <androidx.fragment.app.FragmentContainerView
```
schuetzcarl, line 9: Could you motivate that change ?

schuetzcarl (changes requested):
Thank you for fixing these tests and the for the additional clean up you have made.

I only have two small remarks, then I will agree to merge it.

app/src/main/res/layout/content_main.xml
yanvow: - It is google’s recommended view for hosting fragments.
- Fragment transactions are getting handled by FragmentContainerView under the hood.

app/src/main/java/com/github/factotum_sdp/factotum/MainActivity.kt
yanvow: It's the way to have a navController when you have a FragmentContainerView

app/src/main/res/layout/content_main.xml
schuetzcarl: Ok let's try. It seems to have also some nice animation methods for fragments transition.

schuetzcarl (approved):

PR #60, opened 2023-03-21
search bar added with tests

DanielTavaresA:
- Search bar UI added in route fragement for future use. Will be used for address research.
- Test added to check if Search bar UI is added to screen

### DIFF
--- a/app/src/main/res/layout/fragment_routes.xml
+++ b/app/src/main/res/layout/fragment_routes.xml
@@ -3,35 +3,52 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:orientation="vertical"
     tools:context=".ui.maps.RouteFragment"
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/fragment_routes.xml
```diff
@@ -3,35 +3,52 @@
...
-    <Button
-        android:id="@+id/button_next"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:text="@string/next"
-        android:visibility="invisible"/>
-    <Button
-        android:id="@+id/button_run"
-        android:layout_width="wrap_content"
+    <androidx.appcompat.widget.SearchView
```
yanvow, line 11: You should make the search bar clickable area wider.
A good suggestion would be to use add app:iconifiedByDefault="false"

yanvow (changes requested):

yanvow:
Good code overall with good tests.
However, I have a  remark on the UI/UX of the search button.

yanvow (approved):
Perfect, all good for me

PR #62, opened 2023-03-22
Feature/display proofphoto

jules552:
Basic way for displaying all the photos stored in a Firebase Storage instance. I think the design is ugly for the moment and should be adjusted when implementing seperation of roles in a list of file like a contact list and when you click, actually download the image related.
But for the moment I think it's enough and works pretty well, the image should be cached thanks to FirebaseUI Storage, and when needed, you can refresh the list of photos to check if there is anything new.

I think we should merge it into the main branch


app/src/main/java/com/github/factotum_sdp/factotum/ui/display/utils/PhotoDiffCallback.kt
```diff
@@ -0,0 +1,15 @@
+package com.github.factotum_sdp.factotum.ui.display.utils
+
+import androidx.recyclerview.widget.DiffUtil
+import com.google.firebase.storage.StorageReference
+
+// Callback to calculate the difference between two photo items
+class PhotoDiffCallback : DiffUtil.ItemCallback<StorageReference>() {
+    override fun areItemsTheSame(oldItem: StorageReference, newItem: StorageReference): Boolean {
+        return oldItem.path == newItem.path
```
TimVW7, line 9: Probably can easily implement a test, that checks whether two photo items are the same or not.

TimVW7 (approved):
Overall very well written code, very modular and clean. As you've said, UI maybe isn't super clean yet, but functionality seems to work.

PR #67, opened 2023-03-24
Feature/string address research

DanielTavaresA:
- Adds the possibility to do a research from a string

For now only shows a Snackbar with the print of the address if address valid. Else says that has no found a valid address. In the future result from search will be stored and used.


app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/route/RouteFragmentTest.kt
```diff
@@ -80,4 +90,36 @@ class RouteFragmentTest {
...
+        } else{
+            val bestAddresses = geocoder.getFromLocationName(city, 1)
+            result = bestAddresses?.get(0).toString()
+            onView(withId(com.google.android.material.R.id.snackbar_text))
+                .check(matches(withText(result)))
+        }
+    }
+
+    @Test
+    fun wrongSearchShowsNoResultSnackbar(){
```
jules552, line 119: This test is not passing for and I have this error :

com.github.factotum_sdp.factotum.ui.route.RouteFragmentTest > wrongSearchShowsNoResultSnackbar[Pixel_6_Pro_API_33(AVD) - 13] FAILED 
        androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: view.getId() is <2131231211/com.github.factotum_sdp.factotum:id/snackbar_text>
        If the target view is not part of the view hierarchy, you may need to use Espresso.onData to load it from one of the following AdapterViews:android.widget.ListView{d954126 VFED.VC.. ........ 0,336-1440,1612 #7f08012a app:id/list_view_routes aid=1073741860}
Tests on Pixel_6_Pro_API_33(AVD) - 13 failed: There was 1 failure(s).

Tried two times and same restult

jules552 (changes requested):
Overall pretty good jobs, the code is clean and the functionality seems to work, I have one other complain, is it normal that the snackbars are appearing two times, like one time really fast and then reappear ?
I let you look at it and check if you can fix this two point

DanielTavaresA:
> Overall pretty good jobs, the code is clean and the functionality seems to work, I have one other complain, is it normal that the snackbars are appearing two times, like one time really fast and then reappear ? I let you look at it and check if you can fix this two point

I am not sure about the snackbar appearring two time. Tried on API 33 on my phone and API 30 on emulator and snackbar always appears only once. Could you elaborate how you had this issue ?

jules552 (approved):
Everything has been solved with your latest pushes, and what you added seems ok to be, I agree to merge it!

PR #68, opened 2023-03-27
Feature/sharing outside proofphoto

jules552:
Extending the idea of using the sharing feature to send the link to the image showing the proof photo of a delivery. The button for sharing is linked to an image such that it has access to the information of that image but resulting in an horrible UI in our case. Of course it will be changed when reworking on the interface and role in later sprint.
It is difficult to test for Whatsapp since installing on the CI and setup an account is difficult, same for SMS, only the firing of the intent to share is easily testable I think? 
I also tried implementing for telegram but apparently you need the pseudo and not the phone number so I don't know if it's a good option to put it or not? Even if you still have of course the possibility to share to any other app, but it will not open and put automatically the text.

I put it as a draft for the moment is some want to test it.

### DIFF
--- a/app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/display/DisplayFragmentTest.kt
+++ b/app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/display/DisplayFragmentTest.kt
@@ -1,6 +1,7 @@
 package com.github.factotum_sdp.factotum.ui.display
 
 import android.content.Context
+import android.content.Intent
 import androidx.fragment.app.testing.FragmentScenario
 import androidx.fragment.app.testing.launchFragmentInContainer
 import androidx.recyclerview.widget.LinearLayoutManager
@@ -8,6 +9,8 @@ import androidx.recyclerview.widget.RecyclerView
 import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.action.ViewActions.click
 import androidx.test.espresso.assertion.ViewAssertions.matches
+import androidx.test.espresso.intent.Intents
+import androidx.test.espresso.intent.matcher.IntentMatchers.hasAction
 import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
 import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.ext.junit.runners.AndroidJUnit4
@@ -38,11 +41,13 @@ class DisplayFragmentTest {
         scenario = launchFragmentInContainer(themeResId = R.style.Theme_Factotum)
         Firebase.storage.useEmulator("10.0.2.2", 9199)
         context = InstrumentationRegistry.getInstrumentation().context
+        Intents.init()
     }
 
     @After
     fun tearDown() {
-       emptyStorageEmulator(Firebase.storage.reference)
+        emptyStorageEmulator(Firebase.storage.reference)
+        Intents.release()
     }
 
     @Test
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/factotum_sdp/factotum/ui/display/DisplayFragment.kt
```diff
@@ -30,7 +30,9 @@ class DisplayFragment : Fragment() {
         _binding = FragmentDisplayBinding.inflate(inflater, container, false)
 
         // Set up the recycler view with a photo adapter
-        val photoAdapter = PhotoAdapter()
+        val photoAdapter = PhotoAdapter() { storageReference ->
+            shareImage(storageReference, "1234567890")
+        }
```
DanielTavaresA, line 35: Prop phone number for now, but could be a const to be more clear

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/display/DisplayFragmentTest.kt
```diff
@@ -93,4 +98,23 @@ class DisplayFragmentTest {
...
+            uploadImageToStorageEmulator(context, imagePath, "test_image1.jpg")
+        }
+
+        Thread.sleep(WAIT_TIME_INIT)
+
+        onView(withId(R.id.refreshButton)).perform(click())
+
+        Thread.sleep(WAIT_TIME_REFRESH)
+
+        onView(withId(R.id.shareButton)).perform(click())
```
DanielTavaresA, line 115: If possible the use of thread.sleep should be avoided. Is there a possibility to wait until a new image appears instead ?

DanielTavaresA (commented):
Overall good code. Implementation seems to work as it should. Some comments left for details that could be changed but overall LGTM

app/src/main/java/com/github/factotum_sdp/factotum/ui/display/DisplayFragment.kt
```diff
@@ -54,4 +56,39 @@ class DisplayFragment : Fragment() {
...
+
+            // Create a chooser intent with the option to share via general sharing options and WhatsApp
+            val chooserIntent = Intent.createChooser(generalShareIntent, getString(R.string.share)).apply {
+                putExtra(Intent.EXTRA_INITIAL_INTENTS, arrayOf(whatsappIntent, smsIntent))
+            }
+
+
+            startActivity(chooserIntent)
+        }.addOnFailureListener {
+            // Handle failure here
```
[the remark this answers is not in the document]
jules552: I removed it, since it shoulds works all the time and if not, I don't even now what to display and what could be the cause honestly ?

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/display/DisplayFragmentTest.kt
jules552: Thread.sleep should be avoided in main code, it's not a issue in test code ^^

app/src/main/java/com/github/factotum_sdp/factotum/ui/display/DisplayFragment.kt
jules552: Renamed it to a constant, thanks for the suggestion !

app/src/main/java/com/github/factotum_sdp/factotum/ui/display/DisplayFragment.kt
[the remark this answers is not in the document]
DanielTavaresA: Maybe just a log or snackbar would be enought for the client to understand why it failed (even thought if it fails it will be explicit enought)

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/display/DisplayFragmentTest.kt
DanielTavaresA: I see your point and I agree with you. Most importantly test should be testing if a functionnability is working (which it is the case here). Therefore I can agree to keep this Thread.sleep as having a functionnality tested is more valuable than change the test at this point.

PR #73, opened 2023-03-29
Feature/gps sensor

DanielTavaresA:
- Added the permission to use the current location on the maps view. By allowing to use the current location we now have a point that shows where we are. In the future we could use this point to start an itinerary or share locations to boss/client.

Permissions are hard to test because of revokation of permission is hard. For now I just tested the case where we allow the location to be used


app/src/main/java/com/github/factotum_sdp/factotum/ui/maps/MapsFragment.kt
```diff
@@ -47,11 +59,18 @@ class MapsFragment : Fragment() {
         super.onViewCreated(view, savedInstanceState)
 
         setMapProperties()
+
     }
 
     private fun setMapProperties() {
```
yanvow, line 65: This method is becoming a bit too long and could be refactored into smaller methods.

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/maps/MapsFragmentTest.kt
```diff
@@ -19,20 +20,32 @@ import androidx.test.uiautomator.UiDevice
...
 import java.util.concurrent.TimeUnit
 
 
 @RunWith(AndroidJUnit4::class)
+//used to force first the tests about permissions
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
 class MapsFragmentTest {
 
+    val device = UiDevice.getInstance(getInstrumentation())
+    val buttonTextAllow = when(Locale.getDefault().language){
```
yanvow, line 45: Isn't there another way of getting the "Only this time" for any language?
This is hardcoded for only two languages and limits the test.

yanvow (changes requested):

yanvow:
Overall nice code, working fine when testing it on my end. Moreover, the test coverage is perfect.
The way you forced some tests to be executed first is elegant. The modern way to request permissions was used with `registerForActivityResult()`. And in the `checkAndActivateLocation()` method both ACCESS_FINE_LOCATION and ACCESS_COARSE_LOCATION permissions are check, which is a best practice for handling location permissions.
Some minor changes can be made for more readability.

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/maps/MapsFragmentTest.kt
DanielTavaresA: I understand but I didn't found a way to automate this. And using a translation isn't reliable (ie Deepl has translated "Only this time" -> "Mais cette fois-ci"). For now english and french seem to be the only needed languages. But I agree that this test should be adapted if more languages would be involved

yanvow: I will add it in the backlog

yanvow (approved):

PR #75, opened 2023-03-29
Stores results on file + tests

DanielTavaresA:
- Added the possibility to create a Location and store it into a file in cache. 

For now, no method to retrieve has been done, and to create a location from file is avaliable but should be added in future sprints

### DIFF
--- a/app/src/main/java/com/github/factotum_sdp/factotum/data/localisation/Location.kt
+++ b/app/src/main/java/com/github/factotum_sdp/factotum/data/localisation/Location.kt
@@ -6,6 +6,7 @@ import android.location.Geocoder
 import android.os.Build
 import android.os.Build.VERSION_CODES.TIRAMISU
 import androidx.annotation.RequiresApi
+import java.io.File
 import java.util.concurrent.CountDownLatch
 
 /**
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/factotum_sdp/factotum/data/localisation/Location.kt
```diff
@@ -18,11 +19,49 @@ import java.util.concurrent.CountDownLatch
...
+         * Stored as : addressName CACHE_FILE_SEPARATOR latitude CACHE_FILE_SEPARATOR longitude
+         *
+         * @param query : String. Location to add
+         * @param context : Context. Context in which this method is called
+         * @return returns the location that has been added to the file, or null if no address was found
+         */
+        fun createAndStore(query: String, context: Context): Location?{
+            val location = Location(query, context)
+            if(location.address == null){
+                return location
```
TimVW7, line 37: For me this does not return null if no address was found rather a Location object with a null address. If that's what you wanted maybe updated your description, otherwise return null so it's a bit clearer.

TimVW7 (approved):
Very nice feature, well implemented. The tests seem to cover everything that needs to be tested. Good job.

app/src/main/java/com/github/factotum_sdp/factotum/data/localisation/Location.kt
DanielTavaresA: Good comment thank you !

PR #91, opened 2023-03-30
Feature/research bar contacts

TimVW7:
Search bar works, however it implements a search view, which is practically impossible to test. It might be a good idea to change this to a text view so that we are able to do more tests. I was only able to perform basic tests.


schuetzcarl (approved):
You have found a good solution to implement that search bar on the top of your Contact/ContactsList architecture. 
Concerning the code and the tests they are pretty well overall, just need a quick refactor in one method which is a bit too long. But as we are short in time, and it's not blocking anything I already agree that you merge.

PR #100, opened 2023-04-01
Implement sign up UI

yanvow:
Account creation UI with:
- User name
- Email
- Password
- Role


schuetzcarl (changes requested):
Nice work again, very clean code and test cases well covered.

One important point mentioned in one of my review comment concerns the shared logic and code between the login part and the sign up part.

As it is not possible that a username or a password have different acceptance conditions at login time or signup time, and to ensure a better code maintenance for the future, it has to be changed before this merge.

one solution could be to create an 

`abstract class BaseAuthViewmodel(): ViewModel() `
- with one `MutableLiveData<BaseAuthState>` as attribute for the username and the password.
- with one method : `authDataChanged(username: String, password: String)`
which sets the `isDataValid` attribute.

Make `LoginViewModel` and `SignUpViewModel` inherit from this new class and calling `authDataChanged()` at the end of their respective `loginDataChanged()` and `signUpDataChanged()` implementation which sets the possible new fields depending on their concrete context.

Hence, State classes also need to be grouped together :

`BaseAuthState` as a new abstract class 
with three attributes : username, password and isDataValid 

Make the two concrete `LoginFormState` and `SignUpFormState` inherit from the latter.

app/src/main/java/com/github/factotum_sdp/factotum/ui/signup/SignUpViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+            _signupForm.value = SignUpFormState(passwordError = R.string.invalid_password)
+        } else if (role.isBlank()) {
+            //
+        } else {
+            _signupForm.value = SignUpFormState(isDataValid = true)
+        }
+    }
+
+    // A placeholder username validation check
+    private fun isUserNameValid(username: String): Boolean {
```
schuetzcarl, line 29: Following the design of BaseAuthState, this method could be added to the companion object of the SignUpFormState class. But it's less important. I agree to merge and thanks for your changes.

schuetzcarl (approved):
Thank you for your changes. It looks good to me, let's merge it !

PR #102, opened 2023-04-02
Implement user roles and user names

yanvow:
Add the notion of role and username in login.
The roles as well as the user name are retrieved from the profile-dispatch stored on the database.
Each user has a role that is either :

- Boss
- Courier
- Client


DanielTavaresA (approved):
Overall well coded. Code well adapted for Roles addition. Some comments would be useful to uderstand better the log in process.

DanielTavaresA (approved):
All tests pass locally and seem to test well the main features of login and roles. Implementation well modularised.

PR #103, opened 2023-04-02
UI/contact creation

DanielTavaresA:
- Added a fragment UI for contact creation.

Will be linked in the future to the other fragments to register new contacts


TimVW7 (approved):
Looks good ! Just a remark, when you are writing in the details box, when you press return it keeps writing, whereas all the others it goes out of the box. Might be better to keep the consistency here but it's up to you (also easier to use for users).

DanielTavaresA:
> Looks good ! Just a remark, when you are writing in the details box, when you press return it keeps writing, whereas all the others it goes out of the box. Might be better to keep the consistency here but it's up to you (also easier to use for users).

I understand but for me it is a desirable behaviour. As it is a notes field, maybe the user has multiples details to add and therefore having the possibility to add a line by pressing "enter" is desirable. If you are ok with it I will keep it like that

TimVW7:
Sure, that's another way to look at it. Let's keep it like that then, and we'll modify if we feel like the other way is more user-friendly.

PR #104, opened 2023-04-05
Feature/refactor design display proof photo

jules552:
Reworked the UI to see the different proof photos. I think it's way better now, it shows a list with the date of the different file and the hour, you can simply tap on one of the element of the list and it will open and load into your default image viewer the picture.
It also handle correctly now dark mode so the UI is not anymore invisible and terrible when using it.
And I also modified the sorter such that it sorts the image by the date and not anymore using the name of the files. If any of the file name doesn't respect the naming convention, the sorter will attribute the Unix(0) date and place it at the end so at least it doesn't crash and the name displayed on the list of file is "Unknown date and time".

I think it should be merge, it pass the test but I have issues with the contact list, it seems like clicking on different profil cause crash and empty the production db of its contact :/
I haven't touch anything from what he had made, so don't know what could cause this.

Also removed FirebaseUI Storage since it's the default image viewer that handle downloading and displaying the image now so not needed anymore.

[EDIT] The issue concerning the contacts seems to have disappeared, maybe it was a issue with a bad build don't know, but important to keep looking at it.


app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/display/DisplayFragmentTest.kt
schuetzcarl, line 1: A reformat would be welcome in this class, from line 71 each line is spaced from each other.

And the prefix "displayFragment_" in each test name should be avoided.

app/src/main/java/com/github/factotum_sdp/factotum/ui/display/DisplayViewModel.kt
```diff
@@ -8,6 +8,10 @@ import com.google.firebase.storage.StorageReference
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
+import java.text.ParseException
+import java.text.SimpleDateFormat
+import java.util.*
+import kotlin.Comparator
```
schuetzcarl, line 14: Unused import

schuetzcarl (changes requested):
As usual, it's a very clean work, and the resulting UX is really appreciable.

Adding a test that click on a cardView and check that the correct picture is displayed or at least that the current screen has changed (i.e we are not more on the list screen), should resolve the lack of coverage.

As soon as this and my secondary comments are resolved, I will approve the merge.

jules552:
Everything has been corrected as proposed, should be ready to be merged now !

schuetzcarl (approved):
Perfect ! :)

PR #108, opened 2023-04-07
remove : in image path

yanvow:
The ':' character is not supported for file names on Windows


jules552:
Thank for reporting me that, one more reason why to hate windows...
I had to adapt the format in the picture fragment, but also how file are sorted in the display one...
So some small changes here and there.

jules552 (approved):
Nothing to say, just a small fix so

PR #109, opened 2023-04-07
Create a new user

yanvow:
Adds a user in firebase authentication defining an email and a password for the new user.


schuetzcarl (changes requested):
Seems good overall, but I will make a more elaborate review after you have done a local merge in that branch.
Because there is a lot of conflict, with my last PR #66

app/src/main/java/com/github/factotum_sdp/factotum/data/SignUpDataSink.kt
```diff
@@ -0,0 +1,31 @@
+package com.github.factotum_sdp.factotum.data
+
+import com.github.factotum_sdp.factotum.MainActivity
+import com.google.firebase.auth.FirebaseAuth
```
schuetzcarl, line 4: Unused import

app/src/main/java/com/github/factotum_sdp/factotum/ui/auth/BaseAuthViewModel.kt
```diff
@@ -0,0 +1,13 @@
+package com.github.factotum_sdp.factotum.ui.auth
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+
+abstract class BaseAuthViewModel: ViewModel() {
+
+    abstract val _authResult : MutableLiveData<BaseAuthResult<*>>
```
schuetzcarl, line 9: This line has to be removed. Only the public state has to be an abstract val.
The MutableLiveData attribute has to remain private and depends on the subclass implementation.

schuetzcarl (approved):
I already approve the merge, but  it will be nice if you change my two last remarks before the merge.

Thank you for such a nice job, it is a nice set up for the Authentication, every entity is set according to the data-layer pattern : DataSource & DataRepository. Moreover, now, the common part between the SignUp package and the Login package is perfectly refactored, thanks to the abstract classes in the auth package.

app/src/main/java/com/github/factotum_sdp/factotum/data/SignUpDataSink.kt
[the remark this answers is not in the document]
yanvow: I called it DataSink since it's the opposite of a data source.
It's sending the information given by the user to register.
However, when login in you receive information about the user that logged in and this information will be used in the app.

PR #113, opened 2023-04-09
Increase code coverage for Result class

yanvow:
add tests for Result class


schuetzcarl (approved):
Nice job. I agree to merge it.

PR #114, opened 2023-04-09
Feature/nav header display current user

schuetzcarl:
Done in that PR:
- Bind `User` data collected during the login, to the app's `UserViewModel` observed by the NavMenu Header.


schuetzcarl:
@yanvow 
Note that I have also done these following few minor changes in the architecture when linking our two parts.
- Removed the inner data class `User` from `LoginDataSource`, the `LoggedInUserView` data class and renamed the `LoggedInUser` data class as `User`.
- Removed the `loginRepository` from the `LoginViewModel` args as for the moment there is only one construction option i.e the `LoginRepository` is forced to use the `LoginDataSource` as argument, which itself takes no arguments.

app/src/androidTest/java/com/github/factotum_sdp/factotum/MainActivityTest.kt
```diff
@@ -33,7 +33,7 @@ import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
 
-//Later when non-root fragment will exists : add test for navigateUp
+const val LOGIN_REFRESH_TIME = 3000L
```
yanvow, line 36: This time you used a constant. Well done.

app/src/androidTest/java/com/github/factotum_sdp/factotum/MainActivityTest.kt
```diff
@@ -210,11 +210,20 @@ class MainActivityTest {
...
-        Thread.sleep(3000)
+        Thread.sleep(LOGIN_REFRESH_TIME)
         Espresso.pressBackUnconditionally()
         val uiDevice = UiDevice.getInstance(getInstrumentation())
         assertFalse(uiDevice.currentPackageName == "com.github.factotum_sdp.factotum")
         assertTrue(uiDevice.isScreenOn)
     }
 
+    @Test
+    fun navHeaderDisplaysUserData() {
```
yanvow, line 221: If possible, it might be better to check the correctness of the nav header text variables directly. That way we are sure this text is in the right place.

app/src/main/java/com/github/factotum_sdp/factotum/data/LoginDataSource.kt
```diff
@@ -14,15 +11,13 @@ class LoginDataSource {
...
+    fun login(userEmail: String, password: String, profile: User): Result<User> {
+        val authResultFuture = CompletableFuture<Result<User>>()
 
         auth.signInWithEmailAndPassword(userEmail, password)
             .addOnCompleteListener { authTask ->
                 if (authTask.isSuccessful) {
-                    val loggedInUser =
-                        LoggedInUser(profile.displayName, profile.email, profile.role)
-                    authResultFuture.complete(Result.Success(loggedInUser))
+                    authResultFuture.complete(Result.Success(profile))
```
yanvow, line 20: So there is no more difference between a logged in user and other users, but at least it make this method straight forward.

app/src/main/java/com/github/factotum_sdp/factotum/data/LoginRepository.kt
```diff
@@ -6,12 +6,12 @@ import java.io.IOException
  * Class that requests authentication and user information from the remote data source and
  * maintains an in-memory cache of login status and user credentials information.
  */
-class LoginRepository(val dataSource: LoginDataSource) {
+class LoginRepository(private val dataSource: LoginDataSource) {
```
yanvow, line 9: It's better to restrict dataSource to its usage and make it private, well done.

app/src/main/java/com/github/factotum_sdp/factotum/MainActivity.kt
```diff
@@ -77,18 +77,12 @@ class MainActivity : AppCompatActivity() {
...
-        val user = ViewModelProvider(this, userFact)[UserViewModel::class.java]
+        val user = ViewModelProvider(this)[UserViewModel::class.java]
         binding.navView.findViewTreeLifecycleOwner()?.let { lco ->
-            user.name.observe(lco) {
-                userName.text = it
-            }
-            user.email.observe(lco) {
-                email.text = it
+            user.loggedInUser.observe(lco) {
+                val format = "${it.displayName} (${it.role})"
```
yanvow, line 83: Nice, it's good to display the user role.

yanvow (approved):

yanvow:
Overall, great code and the feature you implemented is tested. 
I will approve the PR even if I think that some additionnal tests can be made.
For example, a test where you login with one user then sign out and login with another user and check if it's the correct name, role and email displayed in the nav header.
Moreover, like I said in my comments, we should probably check if the nav header text variables are correct, if it's possible.
It's up to you if you want to add this test in this PR or in a future one.
Also, thank you for notifying me of the architectural changes in the login system.

app/src/main/java/com/github/factotum_sdp/factotum/UserViewModel.kt
```diff
@@ -1,28 +1,20 @@
...
  * Observed in MainActivity
  */
-class UserViewModel(userValue: String, nameValue: String) : ViewModel() {
-    val email: MutableLiveData<String> = MutableLiveData()
-    val name: MutableLiveData<String> = MutableLiveData()
-    init {
-        email.value = userValue
-        name.value = nameValue
-    }
+class UserViewModel() : ViewModel() {
```
yanvow, line 12: Remove empty primary constructor

app/src/main/java/com/github/factotum_sdp/factotum/data/LoginDataSource.kt
schuetzcarl: Yes not a difference at the Type level, but as finally it is also a `User` instance : in the `UserViewModel`, we have the information by variable naming that the `User` instance stored in is the `loggedInUser`.

schuetzcarl:
@yanvow, I made some of the changes according to your comment, and added a new test scenario which ensures that the displayed user data has changed after a logout.
That scenario works when testing manually, however some Firebase emulator issues make it fails during the connectedCheck.

So for now I masked it from the test bench, you can find it as `private fun` and as the last test of the `MainActivityTest` class. 
The issue is that the second user entry in the login, is not found in the fetched (or not) `usersList`, and hence the snap message : "user not found" is displayed.

I let you resolve that test case if you want in a new Issue/PR according to his priority in that Sprint.
Maybe it is related, but I lose some time on the tests by clearing the auth database manually between each test. If in that same task you could find some code to execute after each or after all tests in a test class, it would be great !

(Last thing in the last commit, I removed the Boolean isProfileRetrieved, as that condition was really perturbing for the UX, it makes the error message "userNotFound" not displayed, and just let the loggin button disabled in case of a problem when fetching the users)

PR #116, opened 2023-04-13
Feature/maps/multiple choice

DanielTavaresA:
- Added the possibility to select address from drop down list in contact creation fragment

Drop down only appears when the search is "submitted". This was a implementation choice as the query is blocking and doing the search for each character was slowing to much the user experience.


app/src/main/java/com/github/factotum_sdp/factotum/data/localisation/Location.kt
```diff
@@ -53,45 +54,54 @@ class Location(query: String, context : Context) {
...
+                tiramisuResultHandler(query, geocoder)
+            } else{
+                resultHandler(query, geocoder)
+            }
+        }
+
+        @RequiresApi(TIRAMISU)
+        private fun tiramisuResultHandler(query: String, geocoder: Geocoder): List<Address>? {
+            var result : List<Address>? = listOf()
+            val latch = CountDownLatch(1)
```
jules552, line 72: So maybe here instead of using a CountDownlatch that will block the main threads and could be the reason why you can't display new result when adding/deleting letter in the search bar, you could use Coroutines that will be using other threads, like in this case maybe the one for IO?

app/src/main/java/com/github/factotum_sdp/factotum/data/localisation/Location.kt
```diff
@@ -53,45 +54,54 @@ class Location(query: String, context : Context) {
...
+                latch.countDown()
+            }
+            latch.await()
+            return result
+        }
+
+        private fun resultHandler(query: String, geocoder: Geocoder): List<Address>? {
+            var result : List<Address>? = listOf()
+            try {
+                val geocodeResult = geocoder.getFromLocationName(query, MAX_RESULT)
```
jules552, line 85: So getFromLocationName is apparently deprecated, no alternative function or way of doing this?

jules552 (changes requested):
Thank you for you work, !

Your feature seems to work as intended, other than my 2 requests I have 2 others remarks, the first one is that it looks bad and not really practical, so maybe I think a redesigned is needed here.
My second remark is more a question, I was wondering if the Location is not a bit too much over-engineered, it seems like there is an API called Google Places API that could handle what you've made, even if you have to create and account an add a billing method but you have free credits each months so for this type of project it should be way enough and I was wondering if it was not simpler?

Other than that I don't see much more to say 👍🏻

app/src/main/java/com/github/factotum_sdp/factotum/data/localisation/Location.kt
DanielTavaresA: The use of the deprecated function is necessary because of the apk used by the CI. The CI is using APK 30, and other functions force APK 33 to be used.

DanielTavaresA:
> Thank you for you work, !
> 
> Your feature seems to work as intended, other than my 2 requests I have 2 others remarks, the first one is that it looks bad and not really practical, so maybe I think a redesigned is needed here. My second remark is more a question, I was wondering if the Location is not a bit too much over-engineered, it seems like there is an API called Google Places API that could handle what you've made, even if you have to create and account an add a billing method but you have free credits each months so for this type of project it should be way enough and I was wondering if it was not simpler?
> 
> Other than that I don't see much more to say 👍🏻

1 - Could you be more precise about what you mean by "looks bad and not really practical" ? 
2 - Google places doesn't seem to be what is needed here. But one API that could be used is Google Geocoding and I could look into to see if it is more useful than the android Geocoding package.

jules552:
> > Thank you for you work, !
> > Your feature seems to work as intended, other than my 2 requests I have 2 others remarks, the first one is that it looks bad and not really practical, so maybe I think a redesigned is needed here. My second remark is more a question, I was wondering if the Location is not a bit too much over-engineered, it seems like there is an API called Google Places API that could handle what you've made, even if you have to create and account an add a billing method but you have free credits each months so for this type of project it should be way enough and I was wondering if it was not simpler?
> > Other than that I don't see much more to say 👍🏻
> 
> 1 - Could you be more precise about what you mean by "looks bad and not really practical" ? 2 - Google places doesn't seem to be what is needed here. But one API that could be used is Google Geocoding and I could look into to see if it is more useful than the android Geocoding package.

For the first point when I say look bad is more about how practical it is. The text seems a bit small and the different option not that well spaced and delimited between each other such that when you tap on it you have no chance to select the wrong option. But it's actually the same issue for the role selection in the sign-up menu so that's why I think you should not care for the moment and we will fix that kind of stuff later

jules552 (approved):
All appropriate changes has been made, you can merge it now !

PR #118, opened 2023-04-17
Feature/archiving drecords

schuetzcarl:
Done in that PR :
- A new `DRecordList` class which decorates a `List<DestinationRecord>`
Really useful to keep most of the previous work done in the `RoadBookViewModel` and `RoadBookViewAdapter` classes, indeed their two main attributes were of type : `List<DestinationRecord>`. 
The decoration adds the necessary API to manage a collection of archived records and for example, to quickly know if a DRecord present in the main List is archived or not. Moreover, it is also possible to switch the main List from a list containing the archived records to a list with none of them. 

- Add the necessary new public functions to the `RoadBookViewModel`
- New `MenuOption` button and button state in the `RoadBookFragment` for showing archived records or not.
- Modify the swipeLeft logic in `RoadBookTHCallback` : 
Only a timestamped record can be archived on a swipeLeft.
Swipe left on a timestamped and archived record ask to unarchive it.
Swipe left on non timestamped record ask to delete it. 

- Add an archived icon displayed on the right of each archived record
- Add test suit


app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordList.kt
```diff
@@ -0,0 +1,90 @@
+package com.github.factotum_sdp.factotum.ui.roadbook
+
+import com.github.factotum_sdp.factotum.data.DestinationRecord
+
+
+private fun displayedRecords(allRecords: List<DestinationRecord>,
```
jules552, line 6: Hum why putting this method outside the DRecordList?

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordList.kt
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Update the current main List (this) in an immutable way,
+     * keeping all the others current settings.
+     *
+     * @param ls: List<DestinationRecord> The new list to replace the current (this) main List
+     * @return a new DRecordList with the main List updated
+     */
+    fun newDisplayedList(ls: List<DestinationRecord>): DRecordList {
```
jules552, line 44: A better name for this type of function could be : `withDisplayedList`

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordList.kt
```diff
@@ -0,0 +1,90 @@
...
+        return DRecordList(allRecs, archived, showArchived)
+    }
+
+    /**
+     * Archive a record
+     *
+     * @param index: Int index of the target DestinationRecord
+     * @return a new DRecordList with the specified record added to the archived records
+     */
+    fun archiveRecord(index: Int): DRecordList {
```
jules552, line 59: Same here `withRecordArchived` coult be a better name

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordList.kt
```diff
@@ -0,0 +1,90 @@
...
+        return DRecordList(allRecords, archived.plus(this[index]), showArchived)
+    }
+
+    /**
+     * Unarchive a record
+     *
+     * @param index: Int index of the target DestinationRecord
+     * @return a new DRecordList containing with the specified record out of the archived records
+     */
+    fun unarchiveRecord(index: Int): DRecordList {
```
jules552, line 69: Same here : `withUnarchiveRecord`

jules552 (changes requested):
Hello and thank you for your work !
From what I have tested your functionalities works well and as intented ! 

The DRecordEditDialogBuilder seems pretty clean to me but if I have to do one small remark is that the functions `waitTimeOrRateFromString`, `timestampFromString`, `actionsFromString` in it could be move outside in a util class since there are not directly related in a functional way to this class.

Second remark, I've seen that sometime you catch really general type of errors like `java.lang.Exception` for example in your `DRecordEditDialogBuilder`. It's a good practice to try catching more specific one since it can help better understanding your code and also possibly not catching other errors that could be thrown by a function but should be treated differently.

Third remark would be about naming, I think most of your class/variables names are great, but sometime you use contraction of word like `DRecord `instead of `DestionationRecord `or `rb` instead of `roadbook`. It's not horrible but it can slowdown the readability and it's often better to have long naming (to a certain extend of course) than trying to compress your variables/class names like that

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordList.kt
schuetzcarl: As there is always a displayedList as main attribute. 
I prefer not to call it with.
I used the "with" prefix for the setting of whether the displayedList is with the archived record or without.

But in that function the displayedList is completely replaced.
Maybe a better name could be `replaceDisplayedList`.

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordList.kt
schuetzcarl: I prefer not as it does not affect the displayedList (i.e the main list of the class directly).

When calling that function you just want to archive a particular record, nothing more.

Then if you want to modify the main list of the DRecordList, the one decorated.
You call the dRecordList.withArchived() which affects the list.

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordList.kt
jules552: Hum I see, I agree with `replaceDisplayedList` then !

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordList.kt
jules552: Then yeah I think put it inside a companion object is the best idea I think !

schuetzcarl:
Thank you for the review.

Yes sure in the `DRecordEditDialogBuilder` during edition or existingEdition I didn't go deep into the Exception error.
But here, as it is local to one of the dialog's edition fields, I thought it was enough to catch any general error coming from there. And I don't want to let pass other kinds of errors coming from that screen.
The cost of going deeper in the error type for each field doesn't worth the candle.

I am sorry for the naming but, I don't know if letting the full name in some cases is less horrible or adds readability, for example : 
DestinationRecordEditBuilder or DestinationRecordDetailsFragment becomes really heavy names.
And yes I sometimes contract more the names when they are variable, if they are obvious, and their type describe the abreviate name. 
For example the variable `rbViewModel` of type `RoadBookViewModel` was an easy contraction, which results in shorter lines throughout the class. But I will try to avoid that in the future, if it really reduces readability.

For your second remark, from what I remembered the parse is specific to the EditDialog, that's why I decided to let it here.

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordList.kt
schuetzcarl: Yes `displayedRecords()` is directly linked to a `DRecordList` and used only by that class internally.

Ok, thank you for revealing that point.

jules552 (approved):
Seems all good to me now !

PR #119, opened 2023-04-18
Feature/display per role

jules552:
Implement different level of visibility of the fragments in the MainActivity depending of the role the user logged-in has.
Also add the ability to redirect the user depending of it's role to a given fragment after getting logged-in.

Bonus : Refactored the Picture to be a bit simpler, not perfect but much better than before.


DanielTavaresA (approved):
Tests seems to check necessary functionnalities. PictureFragment is now a lot more concise and readable which is a good thing. Overall nice work

PR #120, opened 2023-04-18
Feature/update profiles dispatch in firebase after sign up

yanvow:


app/src/main/java/com/github/factotum_sdp/factotum/data/LoginDataSource.kt
```diff
@@ -39,27 +37,27 @@ class LoginDataSource {
         dbRef.child(DISPATCH_DB_PATH).get().addOnSuccessListener {
             if (!it.exists()) {
                 profilesResultFuture.complete(
-                    Result.Error(IOException("Error retrieving profiles"))
+                    Result.Error(IOException("Error retrieving users"))
                 )
                 return@addOnSuccessListener
             }
-            val profileList = it.value as List<*>
+            val profileList = (it.value as Map<*, *>).values.toList()
```
TimVW7, line 44: My compiler doesn't seem to accept this line, should be investigated further, but if it works, on the ci I guess it shouldn't be a problem. It says that it.value being an array list cannot be cast to a map.

TimVW7 (changes requested):
Very nice cleaning of the code, you should change the mainactivitytest import for some of the tests to pass. Otherwise, once everything passes on the ci should be good to go.

app/src/main/java/com/github/factotum_sdp/factotum/data/LoginDataSource.kt
yanvow: I'm changing the way I'm storing profiles in the db.

TimVW7 (approved):
Great, ready for merge !

PR #121, opened 2023-04-18
Feature/set database rules

yanvow:


TimVW7 (approved):
Great, that needed to be done 👍

PR #122, opened 2023-04-19
Feature/link drecord picture

jules552:
Implement the link between the `PictureFragment` and the `DestinationRecord`. So the `PictureFragment` is not anymore accessible with the drawer menu, now it is only inside the `DestinationRecord` when tapping on one of them, you can swap an access it here like previously done, but no it is correctly linked to it so when taking a photo, it will store it locally and inside the Firebase Storage into a file with the name of the client linked to the DestinationRecord.

The feature did not took so much time to implement, but caused a lot of issues in the test with UiSelector mainly but also now have to check what is inside the folder instead of just checking all the files inside the storage. So a lot of rework here and it should be more stable than before!

### DIFF
--- a/app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/picture/PictureFragmentOfflineTest.kt
+++ b/app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/picture/PictureFragmentOfflineTest.kt
@@ -54,20 +54,7 @@ class PictureFragmentOfflineTest {
         device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
         emptyLocalFiles(picturesDir)
 
-        // Open the drawer
-        onView(withId(R.id.drawer_layout))
-            .perform(DrawerActions.open())
-        onView(withId(R.id.roadBookFragment))
-            .perform(click())
-
-        // Click on one of the roadbook
-        val destID = DestinationRecords.RECORDS[2].destID
-        onView(withText(destID)).perform(click())
-
-        // Go to the picture fragment
-        onView(withId(R.id.viewPager)).perform(swipeLeft())
-        onView(withId(R.id.viewPager)).perform(swipeLeft())
-        onView(withId(R.id.viewPager)).perform(swipeLeft())
+        goToPictureFragment()
 
         // Wait for the camera to open
         Thread.sleep(TIME_WAIT_SHUTTER)
--- a/app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/picture/PictureFragmentOfflineTest.kt
+++ b/app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/picture/PictureFragmentOfflineTest.kt
@@ -91,11 +91,11 @@ class PictureFragmentOfflineTest {
 
         storage.reference.listAll().addOnSuccessListener { listResult ->
             fail("Should not succeed")
-        }.addOnFailureListener{
-            //Check if there is a file in the local directory
-            picturesDir.listFiles()?.forEach { file ->
-                assertTrue(file.exists())
-            }
         }
+
+        //Check if there is a folder with a file in it
+        val directories = picturesDir.listFiles()?.filter { it.isDirectory } ?: emptyList()
+        assertTrue(directories.isNotEmpty())
+
     }
 }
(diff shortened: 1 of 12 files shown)


app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/picture/PictureFragmentOfflineTest.kt
```diff
@@ -45,43 +46,55 @@ class PictureFragmentOfflineTest {
...
-
-        // Empty Local Files
         emptyLocalFiles(picturesDir)
 
-        // Launch the fragment
+        // Open the drawer
         onView(ViewMatchers.withId(R.id.drawer_layout))
             .perform(DrawerActions.open())
-        onView(ViewMatchers.withId(R.id.pictureFragment))
+        onView(ViewMatchers.withId(R.id.roadBookFragment))
```
yanvow, line 54: You could import ViewMatchers methods individually so that you don't have to specify `ViewMatchers` every time

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/picture/PictureFragmentOfflineTest.kt
```diff
@@ -45,43 +46,55 @@ class PictureFragmentOfflineTest {
...
+    fun tearDown() {
+        emptyLocalFiles(picturesDir)
+    }
+
     @Test
     fun testDoesNotDeleteFileIfUploadFails() {
         val device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
-        val takePictureButton = device.findObject(UiSelector().description("Shutter"))
-        takePictureButton.click()
+        device.executeShellCommand("input keyevent 27")
```
yanvow, line 78: You should define constants for the input keyevents with explicit names.

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/picture/PictureFragmentOnlineTest.kt
```diff
@@ -45,45 +48,61 @@ class PictureFragmentOnlineTest {
...
-        onView(ViewMatchers.withId(R.id.drawer_layout))
+        // Open the drawer
+        onView(withId(R.id.drawer_layout))
             .perform(DrawerActions.open())
-        onView(ViewMatchers.withId(R.id.pictureFragment))
+        onView(withId(R.id.roadBookFragment))
             .perform(ViewActions.click())
 
+        // Click on one of the roadbook
+        val destID = DestinationRecords.RECORDS[2].destID
```
yanvow, line 60: This code is shared with the PictureFragmentOfflineTest class and could be in a method, so that it's defined only once

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/picture/PictureFragmentOnlineTest.kt
```diff
@@ -45,45 +48,61 @@ class PictureFragmentOnlineTest {
...
+
 
     @Test
     fun testUploadFileCorrectly() {
         val device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
 
-        val takePictureButton = device.findObject(UiSelector().description("Shutter"))
-        takePictureButton.click()
+        // Click on the shutter button
+        device.executeShellCommand("input keyevent 27")
```
yanvow, line 84: The keyevents constants are also shared between the offline and online tests. They could be defined in a Utils class.

app/src/main/java/com/github/factotum_sdp/factotum/ui/picture/PictureFragment.kt
```diff
@@ -15,13 +15,14 @@ import java.io.File
 import java.util.Date
 import java.util.Locale
 
-class PictureFragment : Fragment() {
+class PictureFragment(clientID : String): Fragment() {
```
yanvow, line 18: Could clientID be a UUID instead of a String ?

yanvow (changes requested):
Overall nice code. Moreover, the tests are well adapted to the new usage of the picture fragment. I would sugest some refactoring and modularity changes. 
Finally, I was supprised that when clicking on Picture in the drawer menu the app crashed. From what I understood, you are trying to remove this option in the drawer menu. It might just be an omission.

app/src/main/java/com/github/factotum_sdp/factotum/ui/picture/PictureFragment.kt
jules552: For the moment I dont think so, since theDRecord hold the clientID as a string since it's just the client name, but if we change that or if @schuetzcarl add the UUID in what is stored in the DRecord i'm not against this change, but that will me a small an fast change if we do it so for the moment it's not worth bothering

jules552:
All of your comments have been taken into account except the last for the reasons I've given.
I think it should be ready to be merge into the main branch now

app/src/main/java/com/github/factotum_sdp/factotum/ui/picture/PictureFragment.kt
yanvow: Ok, works for me

yanvow (approved):

PR #124, opened 2023-04-20
can launch navigation from contact details and see destination on map

DanielTavaresA:
- Possibility to see the contact address from the contact details
- Possibility to run directly the route from the contact details

Still need to change the route with the right route when contact creation will be merged

### DIFF
--- a/app/src/main/java/com/github/factotum_sdp/factotum/ui/directory/ContactDetailsFragment.kt
+++ b/app/src/main/java/com/github/factotum_sdp/factotum/ui/directory/ContactDetailsFragment.kt
@@ -40,7 +40,7 @@ class ContactDetailsFragment : Fragment() {
         super.onViewCreated(view, savedInstanceState)
 
         view.findViewById<Button>(R.id.run_button).setOnClickListener {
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/factotum_sdp/factotum/ui/directory/ContactDetailsFragment.kt
```diff
@@ -25,10 +33,26 @@ class ContactDetailsFragment : Fragment() {
...
 
-
         return mainView
     }
 
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        view.findViewById<Button>(R.id.run_button).setOnClickListener {
+            val route = DUMMY_ROUTE[0] //TODO : remove when merged with contact creation and use real route
```
TimVW7, line 43: I completely agree this is a TODO, however since we don't want any TODO comments in our main branch (supposed to be the final product), I would either duplicate the branch and merge this branch without the TODOs or wait for the needed code to come available before merging this branch.

TimVW7 (approved):
Great feature with clean code, what's implemented right now looks great. Missing the correct contact details to fully test it out.

PR #130, opened 2023-04-21
Feature/automatic timestamp

schuetzcarl:
Done in that PR :

- Add `ContextExt`, Extension for the Context of the app to set the permissions more globally. As now `MapsFragment` and the `RoadBookFragment` need to access them.
- Add `LocationClient` and concrete `FusedLocationClient` as source entry for live location data
- Add `LocationService`, which is needed since the last versions of Android limit the localization request per hour, when the app is not displayed on the front-ground. To have a bigger rate, one need to grant the foreground services permissions and to launch explicitly a service.
-  Add a `LocationTrackingHandler` class which handle the front-end logic which interacts with the `LocationService`

- All of this is finally managed in the RoadBookFragment, and the service updates triggers call to the `RoadBookViewModel`'s API to automatically post a new value for the next `DestinationRecord` in the queue (first non-timestamped)

Note that since the back-end with the Client/Contact containing the Location object item that I had to fetch from the `clientID` of a `DestinationRecord`, is not already merged, I did not implemented yet the logic that compute the distance between that object and the Location object given during the `LocationService` updates.
However, everything is ready for that. 
For the moment, in this PR, the next record in the queue is timestamped at each refresh from the LocationService.


app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/LocationTrackingHandler.kt
```diff
@@ -0,0 +1,77 @@
...
+import android.content.Intent
+import com.github.factotum_sdp.factotum.services.LocationService
+import android.content.ComponentName
+import android.content.ServiceConnection
+import android.location.Location
+import android.os.IBinder
+import androidx.activity.ComponentActivity
+import com.github.factotum_sdp.factotum.data.LocationClient
+
+class LocationTrackingHandler() {
```
jules552, line 13: The () of the primary constructor are unnecessary here

app/src/main/java/com/github/factotum_sdp/factotum/ui/login/LoginViewModel.kt
```diff
@@ -41,11 +38,7 @@ class LoginViewModel(private val userViewModel: UserViewModel) : BaseAuthViewMod
                     LoginResult(success = result.data)
             } else {
                 _loginResult.value = LoginResult(
-                    error = if ((result as Result.Error).exception.message == "User not found") {
```
jules552, line 44: Should maybe not be pushed since this modification  purpose is for testing when login is down

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/RoadBookFragment.kt
```diff
@@ -58,6 +62,14 @@ class RoadBookFragment : Fragment(), MenuProvider {
         rbRecyclerView.layoutManager = LinearLayoutManager(context)
         rbRecyclerView.adapter = adapter
 
+        locationTrackingHandler.setOnLocationUpdate {
+            val lat = it.latitude.toString().takeLast(3)
+            val long = it.longitude.toString().takeLast(3)
+            val cal = Calendar.getInstance()
+            println("Location from roadbookFrag: ($lat, $long)")
```
jules552, line 69: It's not really annoying but I guess you don't need anymore the print statement ?

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/RoadBookFragment.kt
```diff
@@ -166,6 +182,31 @@ class RoadBookFragment : Fragment(), MenuProvider {
...
+        val locationSwitch = locationMenu.actionView!!.findViewById<SwitchCompat>(R.id.menu_item_switch)
+        locationSwitch?.let {// Load current state to set the switch item
+            it.isChecked = locationTrackingHandler.isTrackingEnabled()
+        }
+        locationSwitch!!.setOnCheckedChangeListener { _, isChecked ->
+            if(isChecked)
+                locationTrackingHandler.startLocationService(requireContext(), requireActivity())
+            else if(lifecycle.currentState == Lifecycle.State.RESUMED && this.isVisible) {
+                locationTrackingHandler.stopLocationService(requireContext(), requireActivity())
+                println("out from RB")
```
jules552, line 197: Maybe remove this one too?

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/RoadBookFragment.kt
```diff
@@ -231,10 +272,16 @@ class RoadBookFragment : Fragment(), MenuProvider {
         super.onDestroyView()
     }
 
-    private fun fetchRadioButtonState(sharedKey: String, radioButton: MenuItem) {
+    override fun onDestroy() {
+        println("Stop from onDestroy")
```
jules552, line 276: Same here?

app/src/main/java/com/github/factotum_sdp/factotum/services/LocationService.kt
```diff
@@ -0,0 +1,144 @@
...
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+
+private const val CHANNEL_ID = "location_service"
+private const val CHANNEL_NAME = "My Location Service"
+private const val SERVICE_ID = 101
+private const val UPDATE_INTERVAL = 1000L
+class LocationService(): Service() {
```
jules552, line 32: Delete the () since no value passed for constructor

app/src/main/java/com/github/factotum_sdp/factotum/services/LocationService.kt
```diff
@@ -0,0 +1,144 @@
...
+            ACTION_STOP -> stop()
+        }
+        return super.onStartCommand(intent, flags, startId)
+    }
+
+    fun setEventOnLocationUpdate(onLocationUpdate: ((location: Location) -> Unit)) {
+        onLocationUpdateEvent = onLocationUpdate
+    }
+
+    private fun start() {
```
jules552, line 64: Maybe rename this as `startLocationUpdates`

app/src/main/java/com/github/factotum_sdp/factotum/services/LocationService.kt
```diff
@@ -0,0 +1,144 @@
...
+    private fun createNotificationChannel(channelId: String, channelName: String,
+                                          service: NotificationManager): String{
+        val chan = NotificationChannel(channelId, channelName, NotificationManager.IMPORTANCE_HIGH)
+        chan.lightColor = Color.BLUE
+        chan.lockscreenVisibility = Notification.VISIBILITY_PRIVATE
+        service.createNotificationChannel(chan)
+        return channelId
+    }
+
+    private fun stop() {
```
jules552, line 115: Same here as `stopLocationUpdates`

app/src/main/java/com/github/factotum_sdp/factotum/ui/settings/SettingsFragment.kt
jules552, line 1: Empty fragment, should be removed

app/src/main/java/com/github/factotum_sdp/factotum/ui/settings/SettingsViewModel.kt
jules552, line 1: Same here, empty should be removed

jules552 (changes requested):
Thank you for you work ! 

It's a nice start for this feature, some cleaning to do but the overall code seems great to me.
Maybe add a permission request for notification and location when toggleling the feature in the RoadBook. 
And of course as discussed, I think the dependency injection for solving the issue with the CI for the need of google account for the feature to work should be temporary, a proper way of doing this without needing a DI will have to be implement later

app/src/main/java/com/github/factotum_sdp/factotum/ui/login/LoginViewModel.kt
schuetzcarl: Sure, I put the right version back

jules552 (approved):
Seems fine to me now ! 

You should merge it as soon as the CI pass !

schuetzcarl:
Thank you, for the review and sorry for the println garbage, was a bit tired :)

schuetzcarl:
> Thank you for you work !
> 
> It's a nice start for this feature, some cleaning to do but the overall code seems great to me. Maybe add a permission request for notification and location when toggleling the feature in the RoadBook. And of course as discussed, I think the dependency injection for solving the issue with the CI for the need of google account for the feature to work should be temporary, a proper way of doing this without needing a DI will have to be implement later

I took some time to find a way for setting a signed Google user on the CI device but it seems to be not possible. 
The best way to proceed is to change the CI's android API version to 33.

As the GPS sensor is one of the requirement, and the foreground service with FusedLocation can't be launch without a Google Play user authenticated for android API < 33, and as the official documentation specify that "starting from August 2023 New apps must target Android 13 (API level 33)." 
https://developer.android.com/google/play/requirements/target-sdk  
I changed the setting of the CI,  which was quite fast forward, almost no problem happened and we can change it back at any time.

PR #135, opened 2023-04-21
Should hopefully fix some issue causing tests to fails randomly in the CI

jules552:
I think it's pretty descriptive :)


schuetzcarl (approved):
Let's see !

PR #146, opened 2023-04-25
Refactor for ci/roadbook tests

schuetzcarl:
Following #145, where the CI became to much flaky to keep a decent workflow.

I made that PR to make a restart and clean up our tests.

- Reformat and re-indent RoadBookFragmentTest
- Add back two tests from the erased in #145
- Removed in the whole project the SharedPref use which are not safe on the main thread (Migration to DataStore in that sprint)
- Comments some tests in `PictureFragmentTest` and `MapsFragmentTest`, which fails sometimes locally

- Add retry Gradle plugin, which can rerun a failed test immediately without failing the whole build

- Investigate further knowing that our main fail on the CI was : "Test run failed to complete. Instrumentation run failed due to Process crashed. android CI", which happens after a delay, and on a random test, as they have a random execution order.

By adding the Logcat option on .cirrus.yml, I remarked that the cause of all this pain was an automatic update scheduled in the background of the emulator device.
Which, after a delay kill most of the running app in the device, i.e, log line : 
04-24 20:19:09.772   581   607 I ActivityManager: Killing 6639:com.github.factotum_sdp.factotum/u0a160 (adj 0): stop com.google.android.gms due to installPackageLI

I will open an Issue for that, which is the priority know.
After resolving that, one can uncomment, and/or adapt the possible remaining flaky tests. (Or maybe not flaky, just were triggered false by the last explained problem)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -45,6 +45,8 @@ check_android_task:
     ./cc-test-reporter before-build
   screen_record_background_script:
     for n in $(seq 1 20); do adb exec-out screenrecord --time-limit=180 --output-format=h264 - > $n.h264; done
+  start_logcat_background_script:
+    adb logcat > log.log
   check_script: |
     chmod +x firebase
     ./firebase emulators:exec --project sdp-firebase-bootcamp-c795d './gradlew check connectedCheck'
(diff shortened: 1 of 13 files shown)


schuetzcarl:
Going back to the API 30 seems to avoid that problem.
The CI passed three times in a row.

DanielTavaresA (approved):
Ideally tests should be corrected instead of commented. At least having some comments on what is blocking could be useful to decide if those tests should be removed or corrected instead on commented.

schuetzcarl:
Sure but I had enough work with my part. 
And to isolate my flaky ones I needed to do that. 

Now the goal is that you and @jules552 adjust your test in a new specific PR.

Thank you for the approve.

PR #149, opened 2023-04-25
Change login and profile to uid

yanvow:
Change login and profile to uid


app/src/main/java/com/github/factotum_sdp/factotum/placeholder/UsersPlaceHolder.kt
```diff
@@ -57,28 +57,26 @@ object UsersPlaceHolder {
...
-        users.add(USER2)
-        users.add(USER3)
-        users.add(USER_BOSS)
-        users.add(USER_COURIER)
-        users.add(USER_CLIENT)
     }
 
     /**
-     * Populates the database with users.
+     * Populates the database with user.
```
jules552, line 63: I think it's cleaner to user suspendCoroutine to do that type of stuff, like I modified what you did before on my branch like this :

```kotlin
 suspend fun addUserToDb(user: User) = suspendCoroutine { continuation ->
        dataSource.getReference(DISPATCH_DB_PATH).push().setValue(user)
            .addOnSuccessListener {
                continuation.resume(Unit)
            }
            .addOnFailureListener { exception ->
                continuation.resumeWithException(exception)
            }
    }
```
    
Maybe look into it?

jules552 (approved):
Good work ! 

It seems cleaner than before and will be easier to use for permissions in FIrebase, impatient to try it out !
I approve it, but before merging can you just modify what I've suggested?

PR #152, opened 2023-04-26
Refactor/threadsleep to coroutines delay

### LINKED ISSUE #151: Fix CI : background gms update killing our app currently tested
As explained in #146 the CI fails on a random test with an Instrumentation failed error triggered by the device manager who chose to stop our app in order to update the gms package launched in the background.

schuetzcarl:
Try to fix #151 by removing the Thread.sleep() calls in our tests.


schuetzcarl:
Should fix the CI for the background gms update. 
The two times the CI fails during my work, the logcat didn't mention the killing of our app due to the gms update, but the new Firebase FATAL Exception, maybe related to this morning's new Firebase bug on CI.

- I removed all the non necessary Thread.sleep() calls in the project, if they were necessary, I replaced them by a delay() call on a runBlocking{} coroutines.

- Finally, I wrapped the test were a blocking action is done with runTest{} which has the following behavior according to the documentation : 

On JVM and Native, this function behaves similarly to runBlocking, with the difference that the code that it runs will skip [delay](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/index.html)s. This allows to use delay in without causing the tests to take more time than necessary. On JS, this function creates a Promise that executes the test body with the delay-skipping behavior.

jules552 (approved):
We've talked a lot about everything off this PR. Nothing to say and it seems to improve a lot the CI success rate

PR #153, opened 2023-04-26
Feature/client display view

jules552:
Should implement basic seperation between a clients of the display proof pictures section. Now each client only see what they have in their respective folder in Firebase Storage. It's breaking the view of the Boss that should see nothing and should be address in my next task. I also temporary removed the section for the courier.
The feature is quite short, but fixing the tests related to them was a nightmare... The tests should be way faster now, around 45sec faster I would say. 
I decided to remove the snackbar in case of a successful login since it was hiding my refresh button and had to wait 4sec for each test for an unecessary feedback (the snackbar is still here in case of bad outcome during login).
Also for the moment the name of the folder generated is the name of the client (`clientID` in the `RoadBook`), it can be easily modified if we want to replace it with the UUID later.
I've implemented a function to add a user to the database for the emulator in the GeneralUtils object.
I've also fixed the MainActivityTest to be better reliable with the emulator.
Removed all the tests of the LoginFragment. Some of the test like having an account only in the auth and one only in the db cause a lot of shit to happen and I don't see the point of this tests since if it would happen in our production app, we would be screwed honestly since it seems to cause a lot of trouble. So @yanvow will have to rewrite the test, I think other than the 2 I've speaked of, the other should be fine but I prefer for the moment to remove every test and let him look at it.
I solved the issues with the `Picture` tests, was also a nightmare, but the CI should be a bit more reliable than before.
Passed 2 time the CI in a row so doesn't seem to have any major issue left.


schuetzcarl (approved):
Excellent work, another fight against the CI that you won.

You responded to me in person to my two main remarks about that PR.

The first one was about the choice of launch{}.join() instead of runBlocking{}.
Which you tell me that it was a more convenient syntax for you and closer to the other languages.

The second was a warning about why I wrapped my delay() around a runBlocking{} and not letting them in the runTest{} scope (like you did in that PR).
Even if it seems not logical, it is because in practice, in some tests, the delay() were not taking into account, until I it in a new runBlocking{}.  (i.e RoadBookFragmentTest.dragAndDropByInjectionIsWorking())

However, as it seems to work, we can avoid that move here.
Hence, I approve the merge, thank you for the work.

PR #154, opened 2023-04-27
Fix/fixing ci gms

jules552:
Around 15 hours to find a fix for the CI, but seems to have found it.
Swap the image for API 33 but also removed the unnecessary google play store that was included with it since google api is enough in our use case.
Upgrade gradle from 7.5.0 to 8.1.1, build time was around 5min, now around 4min.
Updated some dependencies and removed some also notable one beeing mockk wich is apparently known to cause trouble and random crash in CI.
Also removed one file from @TimVW7 tests in utils to `FillDatabase` or something like that, was doing literally nothing, just crashing the CI because of late init variable not initialized before use.


schuetzcarl (approved):
First of all, thank you for spending some time on that CI problem, it may be really useful for the whole team during the end of the semester.

I have honestly not enough background to review some of the script instructions, but let's see how it will behaves by merging it in our main branch !

Finally, thank you for rebasing your commits, as it would be easier to revert the changes in the worst case.

PR #157, opened 2023-04-28
Migration/sharedpref to datastore

schuetzcarl:
Done in that PR :

- Refactor the `data` package organisation, by adding the `models` and `repositories` packages
- Add `SettingsFragment` which will holds the general settings of our app, and loaded from device DataStore data during the MainActivity onCreate()
- Add other repositories and data class models to perform the feature implementation
- Test suit for that feature

To use the saved RoadBook preferences, one need to set the corresponding check box in the Application Settings fragment before. Otherwise defaults values are set.


DanielTavaresA (approved):
Very nice code and very well implemented. As a user experience I find it not so intuitive to have a fragement with only a field asking the user if he wants to keep his preferences saved (which seems to be the normal behaviour) but current behaviour is also very clear and gives the user more power. One idea could also be to move the Roadbook preferences buttons to app settings fragment and create sections for each fragment that would want to add settings.

schuetzcarl:
Thank you for your remark, I changed the first time ever default preference value to true. Hence, the user don't have to go to the application settings screen to enable a naturally obvious feature.

Yes that was, the idea if the other modules need some general settings, they can add some new fields in the Application settings.

My idea was to keep the RoadBook touch gesture settings in the fragment to be managed more quickly. For example, if the courier want to disable for security the archivage/deletion on a swipe left and enable it back when needed.

PR #159, opened 2023-04-28
Add username to profile dispatch

### LINKED ISSUE #155: Add clientID to user profile

yanvow:


app/src/main/java/com/github/factotum_sdp/factotum/data/SignUpDataSink.kt
```diff
@@ -31,17 +31,33 @@ class SignUpDataSink {
...
-                updateUsersResultFuture.complete(
+                updateUserResultFuture.complete(
                     Result.Error(IOException("Error updating users", it))
                 )
             }
-        return updateUsersResultFuture.get()
+        return updateUserResultFuture.get()
+    }
+
+    fun fetchClientId(clientId: String): Result<String> {
```
TimVW7, line 48: As discussed the feature we now want is to verify if the contacts username matches a contact with no account yet, it is not enough to check if client id already exists.

app/src/main/java/com/github/factotum_sdp/factotum/ui/signup/SignUpFragment.kt
```diff
@@ -54,33 +54,40 @@ class SignUpFragment : Fragment() {
         val emailEditText = binding.email
```
TimVW7, line 54: I think you should replace client id by username and just have it link directly to a contact username. I don't know what you think but since we place on linking contacts and profiles I don't think it is necessary for the user to write his full name in the sign up.

app/src/main/res/layout/fragment_signup.xml
```diff
@@ -86,16 +77,33 @@
...
 
+        <TextView
+            android:id="@+id/textView4"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginBottom="10dp"
+            android:text="@string/clientid_text" />
+
+        <EditText
+            android:id="@+id/clientId"
```
TimVW7, line 98: As I have said previously maybe it would be best to only have a username, email and password field. Up to you, I think if the contact was created previously we don't even need the roles list.

firebase_emulator_data/database_export/sdp-firebase-bootcamp-c795d-default-rtdb.json
```diff
@@ -1 +1,95 @@
...
\ No newline at end of file
+{
+    "contacts": [
```
TimVW7, line 2: smart to format this file a bit 👍

TimVW7 (changes requested):
Overall really clean and modular code. Just add a few changes to incorporate the username being selected from the list of contacts (or in the next task if you think it is better that way). Also your coverage is quite low, consider writing more complete tests that cover the full functionality of your code.

TimVW7 (changes requested):
Great job, code looks really clean and everything seems to work. However please add some more tests. Your coverage is very low so try and get it a tad higher and I'll approve the PR.

jules552:
No time to do it

PR #160, opened 2023-04-29
Bit of cleaning, have to check if it fixes the remaining issue that could still appear

jules552:
Some tests were a bit flakky still, try to address that

### DIFF
--- a/app/src/androidTest/java/com/github/factotum_sdp/factotum/utils/PreferencesSetting.kt
+++ b/app/src/androidTest/java/com/github/factotum_sdp/factotum/utils/PreferencesSetting.kt
@@ -6,13 +6,26 @@ import androidx.test.espresso.Espresso.openActionBarOverflowOrOptionsMenu
 import androidx.test.espresso.action.ViewActions.click
 import androidx.test.espresso.assertion.ViewAssertions.matches
 import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
+import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.espresso.matcher.ViewMatchers.withText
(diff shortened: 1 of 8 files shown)


app/src/androidTest/java/com/github/factotum_sdp/factotum/utils/PreferencesSetting.kt
```diff
@@ -6,13 +6,26 @@ import androidx.test.espresso.Espresso.openActionBarOverflowOrOptionsMenu
...
 object PreferencesSetting {
+
+    private const val WAIT_TIME_BUTTON = 500L
+
     fun enableTouchClick() {
         openActionBarOverflowOrOptionsMenu(ApplicationProvider.getApplicationContext())
-        onView(withText(R.string.rb_label_touch_click)).check(matches(isDisplayed()))
+
+        // Add an explicit wait
+        runBlocking{delay(WAIT_TIME_BUTTON)}
```
schuetzcarl, line 24: Not sure that one is needed, it never fails in DRecordDetailsFragmentTest on the CI. I don't think it fixes some flakyness. Could you get it out ?

schuetzcarl (commented):
Why there is no runTest{}, for exemple for your blocking{} set up or tests in the picture package.

If it's not for a precise reason, maybe it would be nice to add them as it is done uniformly in the project for the tests containing some waiting time.

app/src/androidTest/java/com/github/factotum_sdp/factotum/utils/PreferencesSetting.kt
jules552: One of my test had issue see the "Touch click" button sometime, and it seems to solve it

PR #161, opened 2023-04-29
Replace emulator population during tests to defaults json values load…

schuetzcarl:
…ed before tests run


jules552 (approved):
Not much to say, hope that will result in a more stable/flawless login during test !

PR #163, opened 2023-04-29
Put back Maps tests in a more safe CI

schuetzcarl:


schuetzcarl:
So finally, no need to have the google_apis_playstore on our .cirrus.yml as suggested here : 
https://github.com/sweng-epfl/public/commit/83f51c3c2652848b1ef228198953de23d3fbe179

@DanielTavaresA sorry again for having commented some of your tests the time to set a more reliable CI.
I put them back, and made only one change for clicking on the first route item, in the RouteFragment.
I used a more straightforward operation which may use less UI operations.

However the `b_permissionAllowShowLocation()` still fails locally for me. Maybe I miss a local setting, we can see it together after the next meeting.

Finally, I think, we prefer not to test snackbar if they don't have an important logic, and are a clear app feature.
But if you are sure of the test reliability you can put that test back. It wasn't commented by me.

TimVW7 (approved):
Decommented tests and minor fixes seem to pass tests. Seems good to me !

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/maps/MapsFragmentTest.kt
```diff
@@ -70,51 +71,44 @@ class MapsFragmentTest {
...
+    fun b_permissionAllowShowLocation() {
+        onView(withText(startsWith(FIRST_ROUTE_NAME_PREFIX))).perform(click())
         onView(withId(R.id.button_next)).perform(click())
         if (LocationUtils.hasLocationPopUp()) {
             device.findObject(UiSelector().textContains(buttonTextAllow)).click()
         }
         assertTrue(checkLocationEnabled(testRule))
-    }
+    } */
+
```
DanielTavaresA, line 91: Should be decommented and failing is the correct behaviour as the location mechanism as been disabled unintentionally at some point. I am looking into it to bring it back

DanielTavaresA (changes requested):

DanielTavaresA (approved):
Live location feature added back. All tests are working.

schuetzcarl:
Super !

PR #162, opened 2023-04-29
Revert all major changes made in my client_view branch

jules552:
Title says everything
Just kept the functionality implement and the commented tests causing issues

PR #164, opened 2023-04-30
Put back LoginFragment Test suit

schuetzcarl:
Put back the LoginFragmentTest as it seems that there is no red flags locally to work on the "new" CI version.

It has been removed in #162.


TimVW7 (approved):
Trivial un commenting of tests. Should be good.

PR #165, opened 2023-04-30
Fix/picture display tests

jules552:
All the test except the one `testUploadFileCorrectly` has been restored, I really don't understand how to fix this one, like it's working and not crashing locally but it fails and also crash on the CI...
Also removed one test in the `DisplayFragmentTest` that wasn't important and could cause issue for an unknown reason.

Other than that, everything seems ok, did run 3 times without issue in a row, after merging with the main has to adjust since "Touch click" seems to be activated by default now, so fixed that and everything works like a charm !


app/src/androidTest/java/com/github/factotum_sdp/factotum/utils/LoginMenuIdlingResource.kt
schuetzcarl, line 1: Nice tool that you found from Espresso. :+1:

schuetzcarl (approved):
I agree for your change most of them are keeping the old tests back, but some cleaning and uniformisation has been done. Thank you also for taking into account the new emulator start data loaded before the tests execution.

Just one thing : make a merge squash, on github, hence it would be easier to revert in the worst case.

PR #166, opened 2023-04-30
Automatic Upload Pictures

jules552:
Add capability to the app to schedule automatic upload around every 5 minutes (not precise since Android with schedule in case of business of the phone, battery life, battery mode etc..).

The functionnality cannot really be tested automaticly since it would require to modify the timing of the automatic upload to be quicker during test, but since it's not a precise time like said, we could wait very long time... And other than that, if it does like my test for uploading automaticly when taking the photo and crash during CI test...

So if you want to test by hand, you'll have to wait around 5minutes, sometime less, sometime more, and it seems to anyway upload the file after login. You also have to remove the automatic upload inside the `PictureFragment` or you'll have trouble to test it


app/src/main/java/com/github/factotum_sdp/factotum/MainActivity.kt
```diff
@@ -182,6 +202,7 @@ class MainActivity : AppCompatActivity() {
     companion object {
         private var database: FirebaseDatabase = Firebase.database
         private var auth: FirebaseAuth = Firebase.auth
+        const val INTERVAL_UPLOAD_TIME = 5L
```
yanvow, line 205: Maybe you should precise that these are minutes

app/src/main/java/com/github/factotum_sdp/factotum/ui/picture/UploadWorker.kt
```diff
@@ -0,0 +1,50 @@
...
+        val directoryFiles = directoryEntries.filter { it.isFile }
+
+        return directoryFiles.all { file ->
+            val storageRef = storage.reference.child("${directory.name}/${file.name}")
+            try {
+                storageRef.putFile(file.toUri()).await()
+                file.delete()
+                true
+            } catch (e: Exception) {
+                Log.e("UploadWorker", "Upload error", e)
```
yanvow, line 43: Maybe we should add the name of the file that caused the failure

yanvow (approved):
Overall, the code is functional and appears to work as intended.
I have a comment about a constant name but other then that everything seems good.
I had some ideas about a future version of this upload:

- Consider error handling: The code does handle errors by returning a failure result, but it would be useful to log more specific error messages. For example, if an individual file fails to upload, it would be helpful to know which file caused the error.
- Improve exception handling: The current code catches all exceptions and logs them. It would be better to catch specific exceptions and handle them accordingly.
- Consider adding progress reporting: If there are a large number of files to upload, it would be helpful for the user to see progress updates. We could add a progress bar or some other indication of progress.
- Improve efficiency: The code currently uploads one file at a time. We could use a multi-part upload to upload multiple files in parallel, which can reduce the time it takes to complete the upload process.

jules552:
> Overall, the code is functional and appears to work as intended. I have a comment about a constant name but other then that everything seems good. I had some ideas about a future version of this upload:
> 
>     * Consider error handling: The code does handle errors by returning a failure result, but it would be useful to log more specific error messages. For example, if an individual file fails to upload, it would be helpful to know which file caused the error.
> 
>     * Improve exception handling: The current code catches all exceptions and logs them. It would be better to catch specific exceptions and handle them accordingly.
> 
>     * Consider adding progress reporting: If there are a large number of files to upload, it would be helpful for the user to see progress updates. We could add a progress bar or some other indication of progress.
> 
>     * Improve efficiency: The code currently uploads one file at a time. We could use a multi-part upload to upload multiple files in parallel, which can reduce the time it takes to complete the upload process.


Great idea of parallelizing the upload, haven't fought of that, should work now ! 
I've also fixed both issue you indicated about my code, now the constant says Minute and the error will give the name of the file causing issue.
For the more explicit error to catch and giving some progress and indication to the user, I don't think that's feasible without creating a specific Fragment where the user can see what he has in local file etc... I'm not against this idea but for the moment I think it's important to keep the process hidden from the user

PR #169, opened 2023-05-02
Implement/roadbook offline mode

schuetzcarl:
Done in that PR :

- Add a new `RoadBookRepository` class to manage the different source of data for the RoadBook (i.e remote or local), it also has a cache feature, used when fetching data or for avoiding useless remote or local updates.
- Add a serializers package with a `Serializer` and a `KSerializer` needed for the feature.
- The `DRecordSerializer` is needed for the DataStore read and writes to work.
- The `DateKSerializer` is needed in order to have the @serializable tag on a `DestinationRecord` object, which is useful in the `DRecordSerializer` implementation.

- Add two new settings possibilities for the RoadBook (in the `SettingsFragment`) : 
Now, one can delete the whole RoadBook if needed, and also load the last back-up from the remote, local or application cache source.

At the end of my work I remarked that when de-serializing a remote back-up, hence coming from Firebase, the `timeStamp` attribute is not correctly de-serialized and a default value is used instead.

I spent to much extra time on it without finding a nice solution.
I will open an other Issue for that, it was already enough for that PR.
A solution may be to have our own Timestamp object which would be easier to deserialize. Even if the problem comes only from Firebase's way of serializing/de-serializing objects with the datasnapshot.


schuetzcarl:
I have made the minimal changes to incorporate the shift log into the `RoadBookRepository`, but afterthought, the Shift notion you introduced deserves its own `ShiftRepository` class with its own DB path as you have done in the `DeliveryLogger`.

It will also worth to model a Shift class, as it belongs from our business model : Courier or Boss may have several Shift finished on a same device or application run. For the moment the Shift is at least a User and a final DRecordList, but a timestamp which marks the shift end may be added for example.

Hence, having that model will help you in the new Repository to use DataStore and to manage several Shift data.

DanielTavaresA (approved):
Not a lot to say. Code very well written. Very good modularisation of FirebaseInstance making it more logical to use overall. All functionnalities seem to work well. Only remark would be on the User experience : Having "load last backup" in the settings. doesn't seem to make the most sense to me. Maybe I am missing something here but it looks like giving the courier more work to go to settings fragment to load the backup. We could discuss it together in a meeting to be sure to have the best user experience. But from the code perspective its good.

schuetzcarl:
Thanks for the review.

Ok I take into account your remark.

I put the load here, because for me it is like a rare and an emergency action, and I also didn't want to add it directly on the RoadBookFragment because he is already heavy in features.

However I can move it easily, hence we can discuss about that in the next team meeting.

PR #170, opened 2023-05-02
Hotfix/remove tests

jules552:


TimVW7 (approved):
Useful for making CI, should come back to it later on once you find what the original problem was !

PR #171, opened 2023-05-02
Add login routine to all tests

yanvow:


jules552 (approved):
Not much to review here, a nice addition to finally stop using the work around!

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/roadbook/RoadBookFragmentTest.kt
```diff
@@ -71,7 +72,7 @@ class RoadBookFragmentTest {
     fun toRoadBookFragment() {
         // Ensure "use RoadBook preferences" is disabled
         PreferencesSetting.setRoadBookPrefs(testRule)
-
+        GeneralUtils.fillUserEntryAndEnterTheApp("boss@gmail.com", "123456")
         onView(withId(R.id.drawer_layout))
```
schuetzcarl, line 76: For the RoadBook, the following three lines could be removed, as it is the first Fragment after the login entry.

Also remark that for all other Fragment, a wrong user-entry will still being bypassed by an open action on the DrawerMenu, but I don't think it is a problem for our before routine.

schuetzcarl (approved):
I approve, but check my single comment.
Thank you for updating the whole tests according to the login.

schuetzcarl:
@yanvow, It seems that the new login routine fails on the CI our witness is the RoadBookFragmentTest, having removed the Drawer bypass action seems not working on the CI, even if it succeed locally..

It's probable that for the others FragmentTest classes, it also fails and the Drawer is still used to enter the app.

Maybe for the CI we should let uniformly the Drawer by pass action, and set in all @before the current loggedInUser attribute of the application's UserViewModel. For that, we can add a getter for the UserViewModel in the MainActivity.

Thus, not all the test will depends on the flaky Firebase behavior seen on the CI.

schuetzcarl (changes requested):
Comment needed for the the approval status update.

schuetzcarl (approved):
As time flies and it is needed for #168, I changed it myself for all test classes and it seems to be good.

PR #174, opened 2023-05-03
Feature/location in contacts

DanielTavaresA:
- Add the possibility to have the latitude and longitude in the contact class and store it in the database.


app/src/androidTest/java/com/github/factotum_sdp/factotum/utils/FirebaseStringFormatTest.kt
TimVW7, line 1: I'm not sure the test should be in the utils directory, this directory is more used if other tests need external functions. The test doesn't seem to use any UI either, maybe you can put it directly in the test directory instead of androidTest.

app/src/main/java/com/github/factotum_sdp/factotum/models/Location.kt
```diff
@@ -18,45 +19,58 @@ import java.util.concurrent.CountDownLatch
...
+        val address = geocoderQuery(query, context)?.get(0)
+        coordinates = address?.let { LatLng(it.latitude, it.longitude) }
+        addressName = address?.getAddressLine(0) ?: query
+    }
+
+    constructor(){
+        coordinates = null
+        addressName = null
+    }
+
```
TimVW7, line 36: Very clean way of also being able to create an empty location object, nice !

app/src/main/java/com/github/factotum_sdp/factotum/models/Location.kt
```diff
@@ -84,14 +97,9 @@ class Location(query: String, context: Context) {
         private fun tiramisuResultHandler(query: String, geocoder: Geocoder): List<Address>? {
```
TimVW7, line 97: It's up to you but I think it would be good for these functions to be able to timeout and return with a failure maybe or that it will try again later. Otherwise we might get blocked in the program. Especially with latches.

app/src/main/java/com/github/factotum_sdp/factotum/ui/directory/ContactCreationFragment.kt
```diff
@@ -22,18 +22,19 @@ import androidx.lifecycle.lifecycleScope
...
 
 /**
  * A simple ContactCreation fragment.
  */
-class ContactCreation : Fragment() {
+class ContactCreationFragment : Fragment() {
 
     // Should not stay like that and instead roles should use roles from future ENUM
-    private val roles = listOf("Boss", "Courier", "Client")
+    private val roles = Role.values().map { it.name }
```
TimVW7, line 37: Smart change 👍

TimVW7 (changes requested):
Overall very clean code, very smart of adjusting for incorporating the location in contacts. Just a few things to go over and it's good to go.

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/directory/ContactsUpdateTest.kt
```diff
@@ -173,10 +176,11 @@ class ContactsUpdateTest {
             .check(matches(withText("John")))
         onView(withId(R.id.contact_surname))
             .check(matches(withText("Doe")))
-        onView(withId(R.id.contact_username))
+        val check = onView(withId(R.id.contact_username))
```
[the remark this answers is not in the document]
DanielTavaresA: Seems like a mistake. Don't see why it is there. Thanks for spotting it

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/directory/ContactsCreationTest.kt
```diff
@@ -161,7 +162,8 @@ class ContactsCreationTest {
         onView(withId(R.id.contact_username)).check(matches(withText("@$username")))
         onView(withId(R.id.contact_name)).check(matches(withText("John")))
         onView(withId(R.id.contact_surname)).check(matches(withText("Abby")))
-        onView(withId(R.id.contact_address)).check(matches(withText("123 Main St")))
+        val address = Location.createAndStore("123 Main St", getApplicationContext())
+        onView(withId(R.id.contact_address)).check(matches(withText(address?.addressName)))
```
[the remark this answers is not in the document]
TimVW7: Sorry, just a follow up question, then when we enter an address, in contact creation, that may not be the address used ? So let's say that I write 20 Avenue de Lausanne. Will it enter something like 20 Avenue de Lausanne, Switzerland.

TimVW7 (approved):
Great, all the issues have been resolved, good to go !

PR #175, opened 2023-05-03
Prevent courier from adding contacts

yanvow:
Restrict courier so that they can't add contacts.


TimVW7 (changes requested):
Good code, seems pretty straight forward to me, you should still write a test to verify your functionality works. Otherwise great job.

TimVW7 (approved):

PR #183, opened 2023-05-04
Quick fix boss view

jules552:
Everything is in the title, the files wasn't updating correctly when the boss was swapping between them except if he pressed the "refresh" button explicitly, which is not great.


schuetzcarl (approved):
Nice to you to have seen and fix that before tomorrow's show :)

PR #191, opened 2023-05-05
Feature/clients with sub clients

TimVW7:
Added possibility for clients to have have a super clients.


app/src/main/java/com/github/factotum_sdp/factotum/ui/directory/ContactDetailsFragment.kt
```diff
@@ -55,12 +56,18 @@ class ContactDetailsFragment : Fragment() {
         val contactUsername = view.findViewById<TextView>(R.id.contact_username)
         val contactName = view.findViewById<TextView>(R.id.contact_name)
         val contactSurname = view.findViewById<TextView>(R.id.contact_surname)
+        val contactSuperClient = view.findViewById<TextView>(R.id.managing_client_value)
```
yanvow, line 59: I think you should use bindings

yanvow (approved):
Overall great code bringing a great feature to the app.
I have a little general remark on the non usage of bindings since we use them in the whole app. I think we should align our code.
You are free to change your code directly in this PR or create a new task for it.

PR #194, opened 2023-05-05
Updated such that correct destination is launched from contact detail…

TimVW7:
…s and drecords.

### DIFF
--- a/app/src/main/java/com/github/factotum_sdp/factotum/ui/directory/ContactDetailsFragment.kt
+++ b/app/src/main/java/com/github/factotum_sdp/factotum/ui/directory/ContactDetailsFragment.kt
@@ -17,15 +17,15 @@ import androidx.navigation.findNavController
 import com.github.factotum_sdp.factotum.R
 import com.github.factotum_sdp.factotum.UserViewModel
 import com.github.factotum_sdp.factotum.models.Role
+import com.github.factotum_sdp.factotum.models.Route
 import com.github.factotum_sdp.factotum.placeholder.Contact
-import com.github.factotum_sdp.factotum.placeholder.RouteRecords.DUMMY_ROUTE
 import com.github.factotum_sdp.factotum.ui.maps.MapsViewModel
 import com.github.factotum_sdp.factotum.ui.maps.RouteFragment
 
 class ContactDetailsFragment : Fragment() {
     private lateinit var currentContact: Contact
 
-    private val routeViewModel: MapsViewModel by activityViewModels()
+    private val mapsViewModel: MapsViewModel by activityViewModels()
     private val userViewModel: UserViewModel by activityViewModels()
 
     override fun onCreateView(
@@ -99,16 +99,18 @@ class ContactDetailsFragment : Fragment() {
         }
 
         view.findViewById<Button>(R.id.run_button).setOnClickListener {
-            val route = DUMMY_ROUTE[0] //remove when merged with contact creation and use real route
             val uri =
-                Uri.parse("google.navigation:q=${route.dst.latitude},${route.dst.longitude}&mode=b")
+                Uri.parse("google.navigation:q=${currentContact.latitude},${currentContact.longitude}&mode=b")
             val intent = Intent(Intent.ACTION_VIEW, uri)
             intent.setPackage(RouteFragment.MAPS_PKG)
             requireContext().startActivity(intent)
         }
 
         view.findViewById<Button>(R.id.show_all_button).setOnClickListener {
-            routeViewModel.addRoute(DUMMY_ROUTE[0]) //remove when merged with contact creation and use real route
+            if (currentContact.latitude == null || currentContact.longitude == null) {
+                return@setOnClickListener
+            }
+            mapsViewModel.addRoute(Route(0.0, 0.0, currentContact.latitude!!, currentContact.longitude!!))
             it.findNavController().navigate(R.id.action_contactDetailsFragment2_to_MapsFragment)
         }
     }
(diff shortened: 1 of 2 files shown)


TimVW7:
This is more of a temporary fix, should go deeper as to what to define where.

app/src/main/java/com/github/factotum_sdp/factotum/ui/directory/ContactDetailsFragment.kt
```diff
@@ -119,16 +119,18 @@ class ContactDetailsFragment : Fragment() {
         }
 
         view.findViewById<Button>(R.id.run_button).setOnClickListener {
-            val route = DUMMY_ROUTE[0] //remove when merged with contact creation and use real route
             val uri =
-                Uri.parse("google.navigation:q=${route.dst.latitude},${route.dst.longitude}&mode=b")
+                Uri.parse("google.navigation:q=${currentContact!!.latitude},${currentContact!!.longitude}&mode=b")
```
schuetzcarl, line 123: It would be better to change the currentContact class attribute to not nullable, and handle the null case at the initialisation in onViewCreated(), than to call everywhere .!!. It would add more readability and comprehension for that class.

app/src/main/java/com/github/factotum_sdp/factotum/ui/directory/ContactDetailsFragment.kt
```diff
@@ -119,16 +119,18 @@ class ContactDetailsFragment : Fragment() {
...
             val intent = Intent(Intent.ACTION_VIEW, uri)
             intent.setPackage(RouteFragment.MAPS_PKG)
             requireContext().startActivity(intent)
         }
 
         view.findViewById<Button>(R.id.show_all_button).setOnClickListener {
-            routeViewModel.addRoute(DUMMY_ROUTE[0]) //remove when merged with contact creation and use real route
+            if (currentContact!!.latitude == null || currentContact!!.longitude == null) {
```
schuetzcarl, line 130: For that new latitude and longitude arguments, I remark it now and it should have been implemented in #174,
as we will have that check very often in the whole app, it could be really useful to have an  `hasCoordinates(): Boolean` function in the `Contact` data class

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordDetailsFragment.kt
```diff
@@ -74,8 +81,27 @@ class DRecordDetailsFragment : Fragment() {
             putBoolean("isSubFragment", true)
```
schuetzcarl, line 81: For the different String Navigation argument key, It would be better to write them somewhere as public Constants, as it is used at least two times in the app.

For example, inside the sender class, the class which launch the navigation.

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordDetailsFragment.kt
```diff
@@ -74,8 +81,27 @@ class DRecordDetailsFragment : Fragment() {
...
         }
 
+        var currentContact: Contact? = null
+        try {
+            currentContact = contactsViewModel.contacts.value?.first { c -> c.username == rec.clientID }
+        } catch (e: NoSuchElementException) {
+            // Handle the exception here
+            Log.e("Error", "Could not find contact with ID ${rec.clientID}")
```
schuetzcarl, line 89: Super nice to think about logging a possibly main issue during the execution of the app.

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordDetailsFragment.kt
```diff
@@ -74,8 +81,27 @@ class DRecordDetailsFragment : Fragment() {
...
+        }
+
+        if (currentContact != null) {
+            mapsViewModel.addRoute(
+                Route(
+                    0.0, //Should be the current location
+                    0.0,
+                    currentContact.latitude ?: 0.0,
+                    currentContact.longitude ?: 0.0
```
schuetzcarl, line 98: Here it could be better to display no point on the map when the `currentContact` hasn't some Coordinates set

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordDetailsFragment.kt
```diff
@@ -74,8 +81,27 @@ class DRecordDetailsFragment : Fragment() {
             putBoolean("isSubFragment", true)
         }
 
+        var currentContact: Contact? = null
```
schuetzcarl, line 84: Just a last remark here, consider using contactsViewModel.contacts.value?.let {
 // and write the logic here with a non nullable `currentContact`
}

schuetzcarl (changes requested):
Just some comments about finishing touches concerning the last fixes.

As soon as it is fixed, I agree to merge that. Thank you for your efficient and practical fixes which has a great UX impact.

app/src/main/java/com/github/factotum_sdp/factotum/ui/directory/ContactCreationFragment.kt
```diff
@@ -247,7 +259,7 @@ class ContactCreationFragment : Fragment() {
         return viewModel.contacts.value?.find { it.username == username } == null
     }
 
-    private fun isUsernameExisitingContact(username: String): Boolean {
+    private fun isUsernameExistingContact(username: String): Boolean {
         return viewModel.contacts.value?.find { it.username == username && it.role == Role.CLIENT.name } != null
```
schuetzcarl, line 263: .any() would be more appropriate here. But just for you to know for the next implementations, don't necessary need to modify that here, the null check makes the job.

schuetzcarl (approved):
Perfect ! Let's merge it :)

PR #195, opened 2023-05-05
Bind/automatic timestamp to contact location

schuetzcarl:
Conclude the automatic timestamp after a late receive of #174.

A bit ugly code handling some nullable values which shouldn't be nullable here.

Otherwise added three tests with a mock journey between the "Pelican" and the "Rolex Center".
Still using the `MockLocationClient` for doing that test safely on CI, and being sure to pass it a few hours before the deadline.

However, now `FusedLocationClient` should work on the CI, and will try to find a way to mock the mobile coordinates in the next sprint.


yanvow (approved):
Overall very nice code.
Finally we get this long awaited feature.

PR #198, opened 2023-05-05
Fix/update correctly boss view

jules552:
Baited the fix, should really fix it now...


TimVW7 (approved):
Let's hope the problem is fixed 👍

PR #209, opened 2023-05-06
Offline/delivery log

DanielTavaresA:
- Handles the log of the ended shifts.
- Stores locally if no network is available
- Sends to db once network is available


app/src/main/java/com/github/factotum_sdp/factotum/repositories/ShiftRepository.kt
schuetzcarl, line 1: That class can't use the same logic as the RoadBookRepository, it is an overhead to do that and adds useless operations.

The main difference is that the Shift is collected by an explicit action of the user, and not automatically in the back-ground.

Hence, the first thing is that no cache checking is needed, it has to be stored anyway, when the user click on the end shift button.
Moreover, it should never overrides a previously saved shift. For example, if a user is finishes two shifts offline, both will have to be send to the Database when the network comes back.

So the only "cache" action needed here is to store a collection of shift locally when the network is unavailable and send them upstream when the network comes back. 

Finally, locally if you use a different Date for each shift you will lose the information by using `DataStore<DRecordList>`, using `DataStore<Shift>` instead should makes the job.

schuetzcarl (changes requested):
I am sorry but the `ShiftRepository` class has to be rework. See my comment under that one.

And I did not check the test cases, but they should certainly be adapted accordingly.

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/RoadBookFragment.kt
```diff
@@ -87,9 +92,31 @@ class RoadBookFragment : Fragment(), MenuProvider {
...
+        setShiftRepository()
+
         return view
     }
 
+    private fun setShiftRepository(){
+        if (userViewModel.loggedInUser.value != null){
+            if(rbViewModel.recordsListState.value != null) {
+                Log.i("RoadBookFragment", "Setting ShiftRepository")
+                val currentShift = Shift(
```
schuetzcarl, line 104: This val is unused and it seems to be some debugging log in that function's body.

But that set function should not be used, the repository could be initilaized in the constructor arguments of the `RoadBookViewModel`, which is preferred and aligned to the `RoadBookRepository`.

(An other option is to instantiate it directly at his declaration line.)

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/roadbook/RoadBookFragmentTest.kt
```diff
@@ -386,6 +404,129 @@ class RoadBookFragmentTest {
...
+        assert(ls.zip(shiftList).all { pair ->
+            pair.first.records == pair.second.records
+                    && pair.first.user == pair.second.user
+        })
+
+    }
+
+/*
+    @Test
+    fun shiftIsBackedUpCorrectlyWhenComingBackOnline() {
```
schuetzcarl, line 473: That commented test should be removed before merging

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/roadbook/RoadBookFragmentTest.kt
```diff
@@ -1051,6 +1167,86 @@ class RoadBookFragmentTest {
...
+        }
+        onView(withText(DestinationRecords.RECORDS[index].destID)).check(matches(isDisplayed()))
+        swipeRightTheRecordAt(index)
+        onView(withId(R.id.editTextTimestamp)).perform(click())
+        onView(withText(timePickerUpdateBLabel)).perform(click())
+        onView(withText(R.string.edit_dialog_update_b)).perform(click())
+
+    }
+
+    private fun getRbViewModel(): RoadBookViewModel? {
```
schuetzcarl, line 1182: Prefer using the less possible that getter, it was designed for the Drag & Drop that I could not trigger by simulated touch moves.
However for me the way of testing matters less, and you can let that for the merge.

schuetzcarl (changes requested):
Super clean and efficient rework of the `ShiftRepository` class. It is more straightforward and clear.

I just had some more remarks, as before I didn't made the whole review.

Finally, after those small changes, I am sorry to be a bit strict for that, but 78% could be increased, maybe the commented test does the job.

The main reason is that for that sprints the destabilization tasks, on map will be harder to test, and we need to keep a good coverage on the task we know we can test it.

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/roadbook/RoadBookFragmentTest.kt
```diff
@@ -1051,6 +1116,86 @@ class RoadBookFragmentTest {
...
+            future.complete(shiftDb)
+
+        }.addOnFailureListener {
+            future.completeExceptionally(it)
+            assert(false)
+        }
+        return future.get()
+    }
+
+    private fun setAirplaneMode(state : Boolean){
```
schuetzcarl, line 1182: This function is at the end not used and could be removed

app/src/androidTest/java/com/github/factotum_sdp/factotum/ui/roadbook/RoadBookFragmentTest.kt
```diff
@@ -1051,6 +1116,86 @@ class RoadBookFragmentTest {
...
+    }
+
+    private fun getShiftFromDb(dbShiftRef : DatabaseReference) : List<Shift>{
+        val future = CompletableFuture<List<Shift>>()
+        dbShiftRef.get().addOnSuccessListener { snapshot ->
+            var records : List<DestinationRecord>  = emptyList()
+            var user : User = User()
+            var date : Date = Date()
+            val shiftDb = snapshot.children.mapNotNull { shifts ->
+                shifts.children.forEach {
```
schuetzcarl, line 1158: Did you try here using directly .getValue(Shift::class.java) ? Instead of desalinizing the attributes separately.
It should work and might increase a bit the coverage if the Datastore did not pass in the deserialization part of your ShiftDateKSerializer.

schuetzcarl (changes requested):
Everything looks good in the code :+1: 

Your coverage is sill not high enough, but you really told me that you can't find why.
It seems you will need one additional commit to resolve a conflict, on this one could you try my proposition for the coverage, I just found it in one of your test case.

Then for sure, I will approve the merge, that's obviously a very good job.

schuetzcarl (approved):
You forget to remove `setAirplaneMode()` from the `RoadBookFragmentTest` class, but I already approve the merge.

PR #210, opened 2023-05-06
Cache result of proof pictures

### LINKED ISSUE #180: Try to cache the results (not the images, already handle by picture viewer) of the proof pictures

jules552:
So I've cached the different results of the display fragment so in case of the boss, the different folders, and if accessed, what there is inside is also cached. And for the client, all the results from their folder are cached.
The photos themselves if you click on the cardview are NOT because of limitation of the default image viewer, but the links if you want to share the image via SMS for example are in fact cached.
I also removed the ability to share the image when you're a Client.
Also now the intent to share uses the correct phone number of the client, and if not found, just launch a toast informing the user that the client doesn't have a phone number linked to him.

And to explain quickly how everything works. There is a local db on the phone which is a buffer to all the communications between the display fragment and the Firebase storage. The user will always see what is inside the local db and not what is inside the Firebase storage, the local db will get updated a each request of the user is of course the request is successful.


DanielTavaresA (approved):
Code very well modularised. Easy to understand how hierarchy was built. Could argue that another level of folders based on the day should be added such that not all proofs are on the same page. We can easily imagine that one courier could do at least 10 deliveries per shift. Therefore having all proofs on the same page would become hard to use. Also could argue that having the name of the client delivered in the card would also be useful. If an image cannot be loaded, at least we know to which client it corresponds. But those are all ideas and not fully related to your task which achieves its mean. LGTM

jules552:
> Code very well modularised. Easy to understand how hierarchy was built. Could argue that another level of folders based on the day should be added such that not all proofs are on the same page. We can easily imagine that one courier could do at least 10 deliveries per shift. Therefore having all proofs on the same page would become hard to use. Also could argue that having the name of the client delivered in the card would also be useful. If an image cannot be loaded, at least we know to which client it corresponds. But those are all ideas and not fully related to your task which achieves its mean. LGTM

That's why I want to also add filters and a search bar to better organize a bit, but I will see if I have the time for this type of feature later, thank you for the ideas !

PR #211, opened 2023-05-07
Implement/listen to live location

schuetzcarl:
Prepare some useful tool for @jules552 's task (i.e showing all courier's location on a map)

The main resulting change of that PR is the new observable LiveData fields in the UserViewModel : 
 `userLocation` : to observe the fresh live location of the application user's device.
`userHasTrackingEnabled` : to observe on the activation/deactivation change of the location service.

As an intermediate example, I changed the field mail in the navHeader with the current coordinates.

Other minor things done in that PR :

- Add a `Localisable` Interface, to force the locallisable objects to have coordinates and makes the transition to Android.location.Location object.
- Add two StateFlow read-only attributes to the LocationTrackingHandler.
- Refactor a bit and adjust LocationService, LocationTrackingHandler.
- Move the LocationTrackingHandler from the RoadBookFragment to the UserViewModel for a better scope.

### DIFF
--- a/app/src/androidTest/java/com/github/factotum_sdp/factotum/MainActivityTest.kt
+++ b/app/src/androidTest/java/com/github/factotum_sdp/factotum/MainActivityTest.kt
@@ -162,7 +162,7 @@ class MainActivityTest {
         GeneralUtils.fillUserEntryAndEnterTheApp("boss@gmail.com", "123456")
 
         onView(withId(R.id.drawer_layout)).perform(DrawerActions.open())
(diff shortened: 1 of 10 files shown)


schuetzcarl:
For the moment, I let the `loggedInUser` and the `currentLocation` independent LiveData in the UserViewModel.
As I don't know if and how you will use the User to send an at least (name/location) value to the database.

But everything is ready for both choice with the User implementing Localisable, one can update it with the `withLocation()` method

jules552 (approved):
Honestly everything seems good to me !
Will try to use it asap !

PR #212, opened 2023-05-07
Fix/maps

DanielTavaresA:
- Refactored Location to AddressCoordinates
- Improved handling of query such that search of an address cannot lock
- added the possibility to search from the cache if no network is available 
- added the possibility to create an AddressCoordinates manually (enter addressName and coordinates by hand)
- Changed imported contact values to be coherant with new contact format


app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/RoadBookFragment.kt
```diff
@@ -270,7 +270,7 @@ class RoadBookFragment : Fragment(), MenuProvider {
         endShiftButton.setOnMenuItemClickListener {
             val dialogBuilder = AlertDialog.Builder(requireContext())
             dialogBuilder.setMessage(R.string.finish_shift_alert_question)
-                .setPositiveButton(R.string.end_shift) { dialog, _ ->
+                .setPositiveButton(R.string.end_shift) { _, _ ->
```
schuetzcarl, line 273: Line 283 : an hard-coded String is not registered in the String resources

app/src/main/java/com/github/factotum_sdp/factotum/models/AddressCoordinates.kt
```diff
@@ -0,0 +1,170 @@
...
+            val listToReturn : List<AddressCoordinates>?
+            try {
+                listToReturn = result.get(2L, java.util.concurrent.TimeUnit.SECONDS)
+            } catch (e: Exception) {
+                Log.e("Location", "Error getting location from geocoder",e)
+                return null
+            }
+            return listToReturn
+        }
+/* For now do not need backward handling anymore
```
schuetzcarl, line 112: Large commented code part which should not be merged on the main branch.
If the reason was to don't lose that code, it is not problem because it is kept in the git history.

app/src/main/java/com/github/factotum_sdp/factotum/models/AddressCoordinates.kt
```diff
@@ -0,0 +1,170 @@
...
+            val geocoder = Geocoder(context)
+            if (!isNetworkAvailable(context) || !Geocoder.isPresent()) {
+                return searchInCache(query, context)
+            }
+            return if (Build.VERSION.SDK_INT >= TIRAMISU) {
+                tiramisuResultHandler(query, geocoder, context)
+            } else {
+                Log.e("Location", "SDK < TIRAMISU")
+                return null
+                //resultHandler(query, geocoder, context)
```
schuetzcarl, line 77: That commented line could also be removed

schuetzcarl (approved):
Except from some commented code in `AddressCoordinates` which should be removed before merging, everything looks good and I agree for the merge.

PR #213, opened 2023-05-07
Feature/boss map live location

### LINKED ISSUE #214: Implement a map for the boss to see every active couriers on it

jules552:
Should implement a map that only the boss can access and see the live location of his workers.
Each worker has a color computed using the hash of their uid % 360 since there is only 360 colors available for markers. The map is updated every 30sec with the new location.
Also implement and automatic sending of the couriers' location to a new node of the database every 15sec. Of course you have to activate the location with the roadbook button that does that.
The coverage and in general, the testing is not that great, but honestly I can't really do much more for this kind of feature.


app/src/main/java/com/github/factotum_sdp/factotum/data/LoginDataSource.kt
```diff
@@ -46,7 +46,9 @@ class LoginDataSource {
             val user = User(
                 it.child("name").value as String,
                 it.child("email").value as String,
-                Role.valueOf(it.child("role").value as String)
+                Role.valueOf(it.child("role").value as String),
+                it.child("latitude").value as Double?,
```
schuetzcarl, line 50: Does that modification need to be removed ? I see that a new node is used to send Location, and they are never written on the dispatch. Maybe I'm wrong.

app/src/main/java/com/github/factotum_sdp/factotum/ui/bossmap/BossMapFragment.kt
```diff
@@ -0,0 +1,184 @@
...
+        savedInstanceState: Bundle?
+    ): View? {
+        val view = inflater.inflate(R.layout.fragment_boss_map, container, false)
+        mapView = view.findViewById(R.id.mapView)
+        mapView.onCreate(savedInstanceState)
+        mapView.getMapAsync(this)
+
+        // Initialize Firebase database
+        database = FirebaseDatabase.getInstance().getReference("Location")
+        fusedLocationClient = LocationServices.getFusedLocationProviderClient(requireActivity())
```
schuetzcarl, line 63: It is not needed to launch a second FusedLocationProvider inside the app.
Take the observable in the UserViewModel, you can easily fetch them from here. 

If the live location is not set, a solution would be to use the couriers locations to set the zoom and the camera position, otherwise a default location.

app/src/main/java/com/github/factotum_sdp/factotum/ui/bossmap/BossMapFragment.kt
```diff
@@ -0,0 +1,184 @@
...
+        inflater: LayoutInflater, container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View? {
+        val view = inflater.inflate(R.layout.fragment_boss_map, container, false)
+        mapView = view.findViewById(R.id.mapView)
+        mapView.onCreate(savedInstanceState)
+        mapView.getMapAsync(this)
+
+        // Initialize Firebase database
+        database = FirebaseDatabase.getInstance().getReference("Location")
```
schuetzcarl, line 62: It should at least call getDatabase() to our `FirebaseInstance` object of the app.

But coding all that in the BossFragment does not respect the MVVM pattern we try to respect in our project until there.

schuetzcarl (changes requested):
The implementation indeed works and good solutions have been found for managing upload of the current app user, and fetch of the different locations present in the database.

However, the code doesn't respect the minimal modularization we were respecting until there in the whole project.

See my remarks but in general to much back-end stuff are done in the `BossMapFragment`, you fetch the database data directly from there, launch a new `FusedLocationClient` there, even if there is already an existing one in the project resulting on observable locations in the `UserViewModel`.

Considering the tests, I agree that you can't totally test the feature. But that single test is not enough. 
You can at least check that the values uploaded to the database are the same coordinates you observed inside the app. Then a second test could be after adding one courier location in the database, to check as a boss if exactly one marker is present. You can inspire yourself by the test that Daniel has written for the `MapsFragment`.

Hence for the moment, I don't agree for merging it.

jules552:
> The implementation indeed works and good solutions have been found for managing upload of the current app user, and fetch of the different locations present in the database.
> 
> However, the code doesn't respect the minimal modularization we were respecting until there in the whole project.
> 
> See my remarks but in general to much back-end stuff are done in the `BossMapFragment`, you fetch the database data directly from there, launch a new `FusedLocationClient` there, even if there is already an existing one in the project resulting on observable locations in the `UserViewModel`.
> 
> Considering the tests, I agree that you can't totally test the feature. But that single test is not enough. You can at least check that the values uploaded to the database are the same coordinates you observed inside the app. Then a second test could be after adding one courier location in the database, to check as a boss if exactly one marker is present. You can inspire yourself by the test that Daniel has written for the `MapsFragment`.
> 
> Hence for the moment, I don't agree for merging it.

It should address most of your concern now, just two point I'm not agreeing :

-> `FuseLocationClient` I don't use it simply because from what I've understood, it provides a flow of the current location of the user, here I use a simple `FusedLocationProviderClient` since I just want to know once the current/last position of the boss to provide a city scale zoom on it, I don't really need a flow and wasn't really more conveniant in my case to use it.

-> For testing, I half agree because I just don't know what's gonna be merged with the current Map, what's gonna be scrapped, what's gonna be changed, if we mock the different data provider we are using etc... I think it will more time consumed to do detail testing that are gonna be changed or partially changed than just doing very basic one for the moment and add on them later when proper implementation will be done. It's a bit like the same reason why I haven't modularized every similar part from `MapsFragment` since I don't know what's gonna be kept or not.

Other than that, sorry for the MVVM pattern, I just wanted a rapid working thing and have forgot this kind of stuff, now it should be way better and it should be a better base for the others to work on it now !

schuetzcarl:
Impressed by your fast changes ! Thank you for that.

Concerning the test that is a good argument. I agree with that exceptionally because others may merge with your work or the MapsFragment one, and couls be useful to have it already on the main, to see the differences.  

Could you open a new issue mentioning that test finalization for that task ?

For the FusedLocationProvider, there is indeed a Flow in the LocationTrackingHandler.
But if you use the `userLocation` `LivaData` property of the UserViewModel. You should be able to just make a one time read on it.

However, I think you are right to keep a simple get on the last location as there is the case when the Boss is not working in or navigating to the RoadBook, and doesn't enable at least one time the location. It would be less user friendly to go explicitly there.

But my last remark would be that you could add the method that you call directly from the `FusedLocationProviderClient` to our app implementation : `FusedLocationClient` after adding same method in the `LocationClient` interface. it would be more aligned with the project and will allows a better control for out test strategy. (That I will work on #196)

app/src/main/java/com/github/factotum_sdp/factotum/ui/bossmap/data/BossLocation.kt
schuetzcarl, line 1: A better name could be CourierLocation ? or UserLocation ?
As it is not the boss location, it's a bit confusing.


Also if you could move it to the models package, and making it implements `Localisable`.

We will choose later if it is better for `Localisable` to make the transition to an android.location.Location or some `LatLng`.

But for now just add the latitude and longitude double as argument and construct the `LatLng` when needed on site.
It would be easier in case of serializing etc, and uniform from what we have currently for the `Contact`

schuetzcarl (changes requested):
Just that two final things to change and I will approve the merge.

jules552:
> Impressed by your fast changes ! Thank you for that.
> 
> Concerning the test that is a good argument. I agree with that exceptionally because others may merge with your work or the MapsFragment one, and couls be useful to have it already on the main, to see the differences.
> 
> Could you open a new issue mentioning that test finalization for that task ?
> 
> For the FusedLocationProvider, there is indeed a Flow in the LocationTrackingHandler. But if you use the `userLocation` `LivaData` property of the UserViewModel. You should be able to just make a one time read on it.
> 
> However, I think you are right to keep a simple get on the last location as there is the case when the Boss is not working in or navigating to the RoadBook, and doesn't enable at least one time the location. It would be less user friendly to go explicitly there.
> 
> But my last remark would be that you could add the method that you call directly from the `FusedLocationProviderClient` to our app implementation : `FusedLocationClient` after adding same method in the `LocationClient` interface. it would be more aligned with the project and will allows a better control for out test strategy. (That I will work on #196)

Yes I could have used `FusedLocationProvider` and just collect the first one and have a similar output I think? But like you've said at the end, if you think you can modify it to achieve a similar objective, it's not a difficult change to do honestly, so it's up to you during your work on #196 to maybe swap it with your `FusedLocationProvider` instead of mine.

And thank you for your understanding about the testing issue, I actually think we should use our very good code coverage as a buffer for this sprint, because it's great to have a good coverage but it impacts also the productivity and since it's a heavy sprint due to the destabilization meeting we had, it could be good to be more flexible when we finish building the last features and connects everything, once we have a final prototype, with dependency injection etc... it will be easier to work and test on it since more stable and more representative of the final product and its usage

schuetzcarl:
Indeed, if you want I could do that myself in #196, but take into account my last review comment for `BossLocation`.

app/src/main/java/com/github/factotum_sdp/factotum/ui/bossmap/data/BossLocation.kt
jules552: I renamed them, that was not a good name you're right ! Also moved it into the correct folder and make it implement `Localisable` as asked

schuetzcarl (approved):
Thanks for your cooperation ! I agree to merge it.

PR #216, opened 2023-05-12
Add courier routes in Maps

yanvow:
Add courier routes in Maps:
- Add courier routes (as a straight line between the source and destination of a route) in the MapsFragment
- Routes are clickable
- When clicked, only the clicked route remains
- No tests for this feature (Polylines are not detectable in UiAutomator)


DanielTavaresA (approved):
Works as intedend. But for me, it could have been incorpored with the bigger task that writes the routes in detail as intended in the sprint backlog. I would suggest to merge it in the feature/show-courier-routes-details branch. I let you the choice to which branch you want to merge it as it already implements the focus and drawing a line betweem start and destination.

PR #217, opened 2023-05-12
Bikes on bossmap

### LINKED ISSUE #214: Implement a map for the boss to see every active couriers on it

jules552:
Added the apparently really wanted bikes on the map. At the same time it unlock us RGB colors so we're not locked to the 360 colors from the basic markers in map.
Other than that I swapped the default zoom on the location of the boss, to the geometric median of the couriers' position. Why that ? 2 reasons :

1. It solves the issue with the first zoom not working because acquiring the first time the location can take a bit of delay etc... 

2. It is actually better, you are directly zoomed in to the courier  which has the smallest distance to all the other, so you're pretty much getting zoomed in where the action is happening.

And why the geom median and not the average or the median? 
It's less sensitive to outliers that would move the point in a spot where there is no courier if one or some if the couriers are a bit far away from the rest of the group. 
See https://www.youtube.com/watch?v=iy2RZbq7Kn4 for more info :)

Also updated a bit  the library, deleted unused ones, and updated some of my layout to use Material's ones.

The coverage is still bad, but sinces it's Maps related stuff, not really testable...


yanvow (approved):
Amazing work !
And very useful fixes on top of the actual task.

PR #218, opened 2023-05-13
Fix date deserialization for the roadbook back-up

schuetzcarl:
Fixed the date de-serialization for the roadbook back-up, I was using the wrong parse and format method in the `DateKSerializer`, and was keeping the timing data only.

As aggreed with @DanielTavaresA, I merged the `ShiftDateKSerializer` class with the main as we needed both to have a `Date?` `Serializer` which keep the more possible information about a Date object (date + precise timing).

Finally, the `deserialize` is not traversed in the test, depending on the `DataStore.Serializer` semantic, it is maybe not used. However we are forced to implement it to makes a `DestinationRecord` or a Shift `@Serializable `.


DanielTavaresA (approved):
this PR fixes the serialization problem about the date. Code is adapted where it was needed. LGTM

PR #219, opened 2023-05-14
Implement ClientID check for the Roadbook updates

schuetzcarl:
Done in that PR :
- Add a `clientID` check when the app user edit or create a `DestinationRecord`
The edition is validated only if the `clientID` already exists in the directory (`username` field of a `Contact`).
- Clean and adapt the `RoadBookFragmentTest` class
- Add the necessary contacts in the firebase_emulator_data : QG, More, and New.


TimVW7 (approved):
Nothing much to say. Very clean code, looks good to merge 👍

PR #221, opened 2023-05-15
Feature/see mailboxes as boss

DanielTavaresA:
- Puts mailboxes at the place of delivery destinations. 
- Mailboxes are changed to delivered mailboxes if destination record is timestamped


app/src/main/java/com/github/factotum_sdp/factotum/ui/bossmap/BossMapFragment.kt
```diff
@@ -41,6 +48,10 @@ class BossMapFragment : Fragment(), OnMapReadyCallback {
         }
     }
 
+    private val mailBoxSize = 100
```
jules552, line 51: Maybe move this as a constant like where ZOOM_LEVEL_CITY is

jules552 (approved):
Overall really great job ! I let you merge with the other smalls modifications I've made to the `BossMapFragment` that are present in the main branch and move the constant where I've said and let's go for merging it into the main !

PR #224, opened 2023-05-15
Profile picture UI

yanvow:
- Make the profile picture in the nav header clickable
- You can either select a new profile picuture from your phone photo gallery
- Or view your current profile picture


yanvow:
Unfortunatly, we will not have the time to finish this feature.

PR #225, opened 2023-05-16
Add loggedInUser name to the RB back-up firebase path

schuetzcarl:
- Asked by @DanielTavaresA in today's meeting : that PR add the `loggedInUser` name to the firebase path for the RoadBook back-up.

Needed to dissociates the different RoadBook states on the database according to an app user.

@DanielTavaresA as agreed, I let you change the `BossMapViewModel`, your recent tests and the way you want to read the database to fetch the roadbook for different users.


schuetzcarl:
@jules552 , @DanielTavaresA, the actual merge creates a behavior conflict between your works, but only in `calculateMedianLocation()`, for the moment I made a first adaptation @jules552 you can see it on my last commit.
Feel free to commit if the changes are not appropriated.

As most of the changes from that PR concerns Daniel's one in the BossFragment you created and worked on, I assign you as reviewer.

jules552 (approved):
Seems all good to me !

PR #229, opened 2023-05-17
Feature/see history on mailbox click

DanielTavaresA:


app/src/main/java/com/github/factotum_sdp/factotum/ui/display/DisplayFragment.kt
```diff
@@ -175,19 +175,27 @@ class DisplayFragment : Fragment(), MenuProvider {
...
         }
-
-        val courierBossFolderAdapter = CourierBossFolderAdapter(
-            onCardClick = { clientFolder ->
-                userFolder.value = clientFolder.value
-                observeClientPhotos()
-                setupClientUI()
-            }
-        )
+        val proofPicture = arguments?.getString("ProofPicture")
```
TimVW7, line 178: Maybe change the "ProofPicture" value to a constant, so it may be easier to switch in the future.

app/src/main/java/com/github/factotum_sdp/factotum/ui/maps/BossMapFragment.kt
```diff
@@ -27,19 +32,21 @@ import com.google.android.gms.maps.model.BitmapDescriptorFactory
...
 import com.google.android.gms.maps.model.MarkerOptions
+import com.google.gson.Gson
 
 private const val ZOOM_LEVEL_CITY = 13f
 private const val SCALE_FACTOR_ICON = 0.7f
 private const val FACTOR_DARKER_COLOR = 0.7f
-private const val MAIL_BOX_SIZE = 100
+private const val MAILBOX_TITLE = "Mailbox"
 
 class BossMapFragment : Fragment(), OnMapReadyCallback {
```
TimVW7, line 42: Would be good to add a description since it's a public class.

app/src/main/java/com/github/factotum_sdp/factotum/ui/maps/BossMapFragment.kt
```diff
@@ -130,6 +147,41 @@ class BossMapFragment : Fragment(), OnMapReadyCallback {
...
+                if (bossMapViewModel.history.value != null) {
+                    val jsonStatus =
+                        Gson().toJson(bossMapViewModel.history.value!![client])
+                    Log.d("History", jsonStatus)
+                    bundle.putString("History", jsonStatus)
+                    findNavController()
+                        .navigate(R.id.action_bossMapFragment_to_deliveryHistoryFragment, bundle)
+                    dialog.dismiss()
+                } else {
+                    Toast.makeText(requireContext(), "No history available", Toast.LENGTH_SHORT)
```
TimVW7, line 173: Could be good to add a test case when a contact doesn't have any history.

app/src/main/java/com/github/factotum_sdp/factotum/ui/maps/BossMapViewModel.kt
```diff
@@ -111,10 +116,47 @@ class BossMapViewModel : ViewModel() {
         })
     }
 
+    private fun fetchHistoryState() {
+        logsRef.addValueEventListener(object : ValueEventListener {
+            override fun onDataChange(snapshot: DataSnapshot) {
+                val mapHistoryOfDeliveries = mutableMapOf<String,MutableList<DeliveryStatus>>()
+                snapshot.children.forEach { user ->
+                    user.children.forEach { date ->
```
TimVW7, line 124: Would be good to test this function.

TimVW7 (changes requested):
Really nice code and understandable code. Just a few adjustments and it is good to go. Also the UI for displaying the history could use a bit of work but its good enough for now.

TimVW7 (approved):
Thanks for applying some of the changes, when you have some time, maybe look at the issues from code climate, they might be worth reviewing. Otherwise good to go !

DanielTavaresA:
> Thanks for applying some of the changes, when you have some time, maybe look at the issues from code climate, they might be worth reviewing. Otherwise good to go !

Although I agree with you that code climate issues should be addressed, now the necessity of the feature is worth more of the code quality. Time issues makes it hard for me to address this right now.

PR #230, opened 2023-05-18
Feature/filters display fragment

jules552:
Add a way of filtering the client's photos by selecting a date.
Also at the same time I've fixed the sorting of the photos by date, it was quite a bit broken due to a bad regex.
Also restored the view of the display fragment for the courier and will be similar as the boss one. I think it's too much complicated to setup a third different UI for the courier to see only his photos when he will actually not use this menu that much in my opinion, and since @schuetzcarl  said it's not that much of a privacy concern that courier can see that X client got a package delivered since it's more a business oriented stuff and targetting more small/middle size enterprises. 

Also fixed a small merging mistake made on the boss map, now the bikes should be displayed properly again on the boss map


app/src/main/java/com/github/factotum_sdp/factotum/ui/display/DisplayFragment.kt
```diff
@@ -83,9 +128,9 @@ class DisplayFragment : Fragment() {
                 observeBossFolders()
                 setupBossUI()
             }
-            Role.CLIENT -> {
-                observeClientPhotos()
-                setupClientUI()
+            Role.COURIER -> {
+                observeBossFolders()
+                setupBossUI()
             }
```
TimVW7, line 134: Maybe a better name or a comment would be great when calling this function. It makes it seem as if the when someone has the courier role, he has access to the bossUI

TimVW7 (approved):
Very nice and clean code, feature works great. Maybe try and see if you can add a more exhaustive test but it looks good !

PR #231, opened 2023-05-19
Feature/courier bag

schuetzcarl:
Implement the whole courier_bag feature.

It consist in having a new Fragment on the top of the RoadBook to record at each place if according to the actions planned, we pick or deliver some package : new class `Pack` in the app.

If we pick a new Pack at a certain `DestinationRecord`, then a new Dialog ask us to fill the information about the pack and his recipient. Then the created pack is visible in the new `BagFragment.`

If we have a deliver action at a `DestinationRecord`, the packages in the `BagFragment`  which have the client corresponding to that last DestinationRecord as recipient will be timestamped, and still be displayed in the BagFragment with the marker Delivered at : filled.

Finally the semantic is robust to the possible changes in the  RoadBook, i.e DestinationRecord : deletion, clientID edition, timestamp edition or deletion. 

Note that I will add the onClickListener with the package notes editable TextField, there is already enough to review on that single PR.


app/src/main/java/com/github/factotum_sdp/factotum/ui/bag/BagAdapter.kt
```diff
@@ -0,0 +1,87 @@
...
+        override fun areItemsTheSame(oldItem: Pack, newItem: Pack): Boolean {
+            return oldItem == newItem
+        }
+
+        override fun areContentsTheSame(oldItem: Pack, newItem: Pack): Boolean {
+            return oldItem.packageID == newItem.packageID
+        }
+    }
+
+    companion object {
```
jules552, line 83: This type of strings should be more putted in the values/strings.xml and use getstring().

app/src/main/java/com/github/factotum_sdp/factotum/ui/roadbook/DRecordList.kt
```diff
@@ -118,6 +189,48 @@ class DRecordList(
      * @return a new DRecordList with the main (this) list displaying only the non-archived records
      */
     fun withoutArchived(): DRecordList {
-        return DRecordList(allRecords, archived, false)
+        return DRecordList(allRecords, archived, timestamped, false)
+    }
+
+
+    /** Helpers : */
+    private fun computeTimestampedOnAdd(recordToAdd: DestinationRecord): Set<DestinationRecord> {
```
jules552, line 197: Possible way to refactor this, like you call it computeTimestamped(action_add : boolean) or something like

jules552 (approved):
Overall pretty good code, well tested and documented, some very small changes to maybe do like I've indicated but other than that. The feature seems to work as expected, apparently you fixed the bug we've encountered during the presentation where not filling the info about the package taken leading to not be able to open the bag so perfect !

schuetzcarl:
@jules552 , for the bug I didn't fixed it, I just think that the merge on the demo's branch goes wrong. But I am not sure.

Thank you for the review, I will add the two minor changes you asked in my new PR for the package click, in the next and last Sprint.

PR #232, opened 2023-05-19
Magic to make caching and fetching faster

jules552:
Async parallel fetching to Firebase Storage, also the local db is not in order of date for the photo and name for the folder, so sorting is done when inserting new elements and overall should be more responsive. Same for filtering, it's now done by SQL queries, should be way faster. And tons of micro optimizations resulting in an overall better experience.

Last thing that could be done is pagination, but not possible on Firebase Storage, possible on the local db but apparently complicated without Jetpack Compose so unlucky I guess :(


schuetzcarl:
Could be really useful when using the app in a real context. As the picture proof is an heavy data and also one of that scale the most in our app.

The implementation looks good to me, well done.

schuetzcarl (approved):
Could be really useful when using the app in a real context. As the picture proof is an heavy data and also one of that scale the most in our app.

The implementation looks good to me, well done.

PR #245, opened 2023-05-21
Implement/package notes edit text

schuetzcarl:
Add the package's notes dialog, appearing when clicking on the corresponding package.

Add also an indicator icon that appears on the package item only when it contains some notes.


DanielTavaresA (approved):
Everything works as inteded. I just have one general comment on the bag that is that items cannot be selected when a delivery at a client is done. It could be interesting to have the possibility to select which item is delivered and not have all item automatically selected as removed from the bag. Other than that LGTM great job.

PR #246, opened 2023-05-22
fix to show only current address

DanielTavaresA:
- only shows the delivery address in the map fragment in destination details records

### DIFF
--- a/app/src/main/java/com/github/factotum_sdp/factotum/ui/maps/MapsFragment.kt
+++ b/app/src/main/java/com/github/factotum_sdp/factotum/ui/maps/MapsFragment.kt
@@ -19,6 +19,7 @@ import com.google.android.gms.maps.GoogleMap
 import com.google.android.gms.maps.SupportMapFragment
 import com.google.android.gms.maps.model.LatLng
 import com.google.android.gms.maps.model.LatLngBounds
+import com.google.gson.Gson
 
 
 /**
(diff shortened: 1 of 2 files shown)


DanielTavaresA:
> I unfortunately get a crash when testing manually and clicking on the Maps menu icon with my [enrico@g.com](mailto:enrico@g.com) Boss account. Do you have the same ?

Error seems to be related to Sheet-shift in firebase being inconsistent. If you go look, today sheet-shift records are stored in a strange way having the list of records along with the users node. I don't know how but some versions using the old way of storing the sheet-shift must be used or we are having an issue with it.

schuetzcarl:
It is ok you were right for the DB path.

However during manual testing, I found the the case the client has no address, it display the positions and the routes of the Maps Fragment. Whereas it should display maybe an empty map.
Even if for sure normally no client have no address.

Then in the Maps Fragment, @yanvow should see what he wants, one nice thing could be to display all the location of the RoadBook with their planning number as little text on top. That's simply their index in the RoadBookViewModel's list.

schuetzcarl (approved):
Otherwise concerning the code and the way you implemented, it looks good to me, already before the last commit.

My last comment could be done in another PR, but it's a really important thing to have before Friday, otherwise the Maps fragment for a Courier when launched form the menu is a bit unused.

PR #248, opened 2023-05-22
Fixing Crashing issue DRecordDetails with Picture Fragment

jules552:
I think it should fix this issue, my attempt to reproduce it failed so ?

But I disagree with the idea of the return button not sending back to the roadbook because if you fail your photo and want to retake it, you can in the camera app without needing pressing the back button. So adding a button inside the fragment to retake a photo is not really pleasing solution...
I've tried unsuccessfully to send back the user on the DRecordDetails' first fragment when you tap on it, but it resulted in crashing in all the ways I've tried, my best solution was trying simply to relaunch the fragment with the same destID, but I don't know why it's crashing... So if you have an idea how to do this @schuetzcarl since I suppose it's a issue with the navgraph could you maybe look at it, but I don't think it's really a issue and I don't see anyone really complain about that in the app presentation.


app/src/main/java/com/github/factotum_sdp/factotum/ui/picture/PictureFragment.kt
```diff
@@ -74,7 +108,7 @@ class PictureFragment(clientID: String) : Fragment() {
         if (isGranted) {
             openCamera()
         } else {
-            findNavController().popBackStack()
+            findNavController().navigateUp()
```
schuetzcarl, line 111: Just removing that call line, and doing nothing in the else, let the device going back where the Intent have been launched, without having an Exception.

app/src/main/java/com/github/factotum_sdp/factotum/ui/picture/PictureFragment.kt
```diff
@@ -64,8 +99,7 @@ class PictureFragment(clientID: String) : Fragment() {
             } else {
                 photoFile.delete()
             }
-
-            findNavController().popBackStack()
+            findNavController().navigateUp()
```
schuetzcarl, line 102: That one also.

schuetzcarl (commented):
I marked the two lines to be removed in order to solve the back problem when pressing it from th e picture app of the device.

The only thing needed is an explicit button on your new screen (`FragmentPictureBinding`), with a label "Take proof picture" for example.

If you disagree, we will asks the others at the meeting, but for me it adds a lot of readability, and in the case you don't want to retake a picture but just cancel the picture try, it seems natural to comeback on the `DRecordDetailsFragment`.

jules552:
> I marked the two lines to be removed in order to solve the back problem when pressing it from th e picture app of the device.
> 
> The only thing needed is an explicit button on your new screen (`FragmentPictureBinding`), with a label "Take proof picture" for example.
> 
> If you disagree, we will asks the others at the meeting, but for me it adds a lot of readability, and in the case you don't want to retake a picture but just cancel the picture try, it seems natural to comeback on the `DRecordDetailsFragment`.

Just deleting the line just relaunch the camera when pressing on the back button. And for me the button is a horrible design choice, it's like if you had to press a button to make the map appear in the other menu, or the contact etc.. It's breaking the consistency and add a click for no huge gain

schuetzcarl:
> > I marked the two lines to be removed in order to solve the back problem when pressing it from the picture app of the device.
> > The only thing needed is an explicit button on your new screen (`FragmentPictureBinding`), with a label "Take proof picture" for example.
> > If you disagree, we will asks the others at the meeting, but for me it adds a lot of readability, and in the case you don't want to retake a picture but just cancel the picture try, it seems natural to comeback on the `DRecordDetailsFragment`.
> 
> Just deleting the line just relaunch the camera when pressing on the back button. And for me the button is a horrible design choice, it's like if you had to press a button to make the map appear in the other menu, or the contact etc.. It's breaking the consistency and add a click for no huge gain

It relaunch because you automatically send the Intents on the `onResume()`, having an explicit button or something else, solves that.

The maps tab is a different situation, the Maps screen is contained inside a fragment which itself is inside our app. One can easily move out with the tabs, while staying in the `DRecordDetailsFragment`.

jules552:
> > > I marked the two lines to be removed in order to solve the back problem when pressing it from the picture app of the device.
> > > The only thing needed is an explicit button on your new screen (`FragmentPictureBinding`), with a label "Take proof picture" for example.
> > > If you disagree, we will asks the others at the meeting, but for me it adds a lot of readability, and in the case you don't want to retake a picture but just cancel the picture try, it seems natural to comeback on the `DRecordDetailsFragment`.
> > 
> > 
> > Just deleting the line just relaunch the camera when pressing on the back button. And for me the button is a horrible design choice, it's like if you had to press a button to make the map appear in the other menu, or the contact etc.. It's breaking the consistency and add a click for no huge gain
> 
> It relaunch because you automatically send the Intents on the `onResume()`, having an explicit button or something else, solves that.
> 
> The maps tab is a different situation, the Maps screen is contained inside a fragment which itself is inside our app. One can easily move out with the tabs, while staying in the `DRecordDetailsFragment`.

Yes I know from where the 'issue' come from, but my example was here to say that there is no button to show the map, the detail, the contact infos... You have direct access to those parts, and I want to keep the same flow with the photo. And also adding a button will add one more tap each time when instead in this case, you have to do one more tap just when you fat fingered and wanted to go on another menu maybe, but from the huge distance between the tabs buttons I don't think it will happens often so

schuetzcarl:
> > > > I marked the two lines to be removed in order to solve the back problem when pressing it from the picture app of the device.
> > > > The only thing needed is an explicit button on your new screen (`FragmentPictureBinding`), with a label "Take proof picture" for example.
> > > > If you disagree, we will asks the others at the meeting, but for me it adds a lot of readability, and in the case you don't want to retake a picture but just cancel the picture try, it seems natural to comeback on the `DRecordDetailsFragment`.
> > > 
> > > 
> > > Just deleting the line just relaunch the camera when pressing on the back button. And for me the button is a horrible design choice, it's like if you had to press a button to make the map appear in the other menu, or the contact etc.. It's breaking the consistency and add a click for no huge gain
> > 
> > 
> > It relaunch because you automatically send the Intents on the `onResume()`, having an explicit button or something else, solves that.
> > The maps tab is a different situation, the Maps screen is contained inside a fragment which itself is inside our app. One can easily move out with the tabs, while staying in the `DRecordDetailsFragment`.
> 
> Yes I know from where the 'issue' come from, but my example was here to say that there is no button to show the map, the detail, the contact infos... You have direct access to those parts, and I want to keep the same flow with the photo. And also adding a button will add one more tap each time when instead in this case, you have to do one more tap just when you fat fingered and wanted to go on another menu maybe, but from the huge distance between the tabs buttons I don't think it will happens often so

For me, that additional click completely worth in that situation, It is less surprising to pop on a new external app, and, it is more natural. Using tab as button is quicker but less comprehensive. Also for going back it will be clearer. Let discuss it with the others.

schuetzcarl (approved):
However this fix is sufficient to solve the main crash, hence I approve it.

schuetzcarl:
But just have seen that now : A single test for checking that a picture can be taken from the DRecordDetailsFragment would be welcome for the coverage. Thanks in advance !

PR #249, opened 2023-05-22
Default profile picture

yanvow:
Set the default profile picture (nav header & contact details) to this:
image


TimVW7 (approved):
Sure why not, maybe could differentiate between different contacts (boss, client, courier). Does look better this way.

PR #251, opened 2023-05-23
Implement/timed rb backup

schuetzcarl:


yanvow (approved):
Great code and tests. This feature comes at the right moment since we are finalizing the app.

PR #255, opened 2023-05-24
Implement/bag backup

### LINKED ISSUE #236: Add BagRepository and DataStore for back-up

schuetzcarl:
Done in that PR : 

As the RoadBook, now the Bag packs are backed-up on Firebase or locally if no network is available.
Then, when loading the RoadBook back-up in the application settings, the Bag's one also is loaded.

- Add abstract class `BackUpRepository`
- Add `BagRepository` class
- Replace the runnable routine for the RoadBook back-up by an upload, at each change, as the old onPause() case covers the case when the user is out of the RoadBookFragment.


yanvow (approved):
Great code and tests. The code is well structured and modularized. Also, it adds another offline feature.

PR #256, opened 2023-05-24
added names on markers

DanielTavaresA:


schuetzcarl (changes requested):
It is enough for tomorrow, just remove the red routes lines. 
And then I will approve the PR.

The red lines would give something if they connect each point between them but all coming from the actual localization, doesn't seems relevant.

schuetzcarl (approved):
It is fine now, well done.

PR #257, opened 2023-05-24
Launch phone call when clicking phone number

yanvow:
Testing wasn't possible because you can not hang up the phone after clicking on the phone number.
Test looks like this:
```
@Test
    fun clickOnPhoneNumberOpensCall() {
        init()
        onView(withId(R.id.contact_phone))
            .perform(click())
        intended(hasAction(Intent.ACTION_CALL))
        release()
    }
```


TimVW7 (approved):
Great feature, seems to work as intended. As mentioned very hard to test, so isn't necessary especially considering time constraints.

PR #258, opened 2023-05-24
Implement BagFragment filter

schuetzcarl:
Implemented a default filter on the BagFragment.

When arriving on the fragment, only the non-delivered packs are displayed, like a real bag !
The delivered packs could still be displayed by just clicking on the top-right "send-arrow" icon.


jules552 (approved):
The code seems alright and well tested !
The functionality is working as intended, maybe the only thing I would change is the icon to see the past delivery made, would maybe more use that :

https://fonts.google.com/icons?selected=Material+Symbols+Outlined:history:FILL@0;wght@400;GRAD@0;opsz@48&icon.query=history

Because the one you choose is too similar to a chat/send icon so it can be a bit confusing

PR #264, opened 2023-05-25
Refactor some class in the correct package & Add delete all RoadBook …

schuetzcarl:
…on EndShift click

- Just move some classes, and rename a package.
- Also now, the RoadBook is cleared when the EndShift button is clicked.


jules552 (approved):
Nice cleaning up !

PR #266, opened 2023-05-25
Quick fix to disable conflicts between record details swiping and map swiping.

TimVW7:
Just disabled maps scrolling when in the record details.


schuetzcarl (approved):
It makes the job :)

PR #269, opened 2023-05-26
Get rid of demo records out of the tests

schuetzcarl:
- Get rid of the demo records 
- And clear also RoadBook internal occurrences map when the RoadBook is cleared


jules552 (approved):
Nothing to say, let's merge it

PR #273, opened 2023-05-26
launch phone call when clicking phone number

yanvow:


schuetzcarl (approved):
Welcome back.
Already approved in #257. Was just reverted for CI purpose, but now it is fine.
Hence, I approve it.

PR #276, opened 2023-05-26
Fix/sign in menu button

schuetzcarl:
- Fixed the sign in bypass
- Fixed the location switch which was staying enabled even if the live location was turned off


jules552 (approved):
Hopefully improve the CI and welcoming fix to the app !

PR #279, opened 2023-05-26
Fix/default rb preferences

schuetzcarl:


jules552 (approved):
Nice final fix for our project, well done !

